PR #5, opened 2024-05-14
Add publish github action.

leoromanovsky:
## motivation

automate the publishing step in a cloud environment with a consistent build output

## changes

* adds a github step whenever a tag is created. we will create a github "release" and tag together in one step.
* added `CRATES_IO_TOKEN` to the repository's secrets

Screenshot 2024-05-14 at 10 58 02 AM

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,9 +1,8 @@
 name: Publish Crate
 
 on:
-  push:
-    tags:
-      - '*'  # Trigger when a tag is created
+  release:
+    types: [published]
 
 env:
   CARGO_TERM_COLOR: always
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,30 @@
+name: Publish Crate
+
+on:
+  push:
+    tags:
+        - '*'  # Trigger when a tag is created
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  publish:
+    name: Publish to Crates.io
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        toolchain:
+          - stable
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Rust toolchain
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,30 @@
+name: Publish Crate
+
+on:
+  push:
+    tags:
+      - '*'  # Trigger when a tag is created
```
rasendubi, line 6: minor: I usually prefer triggering release when a GitHub release is created:
```suggestion
on:
  release:
    types: [published]
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Install Rust toolchain
+        run: rustup update ${{ matrix.toolchain }} && rustup default ${{ matrix.toolchain }}
+      - name: Build Release
+        run: cargo build --release --verbose
+      - name: Test
+        run: cargo test --verbose
+      - name: Publish
+        run: cargo publish --token ${{ secrets.CRATES_IO_TOKEN }}
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CRATES_IO_TOKEN }}
```
rasendubi, line 30: nit: I believe we only need one way to pass the token
```suggestion
        run: cargo publish
        env:
          CARGO_REGISTRY_TOKEN: ${{ secrets.CRATES_IO_TOKEN }}
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,30 @@
...
+        toolchain:
+          - stable
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Rust toolchain
+        run: rustup update ${{ matrix.toolchain }} && rustup default ${{ matrix.toolchain }}
+      - name: Build Release
+        run: cargo build --release --verbose
+      - name: Test
+        run: cargo test --verbose
```
rasendubi, line 26: minor: let's check that documentation builds as well
```suggestion
      - name: Test
        run: cargo test --verbose
      - name: Docs
        run: cargo doc --verbose
```

rasendubi (approved):

PR #6, opened 2024-05-15
chore: dump MSRV from 1.76.0 down to 1.65.0

rasendubi:
Specify Minimum Supported Rust Version and dump it down to 1.65.0.

Best reviewed by commit. (Also better to merge or rebase instead of squashing.)

### DIFF
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -1,0 +1,23 @@
+name: Verify MSRV version
+
+on:
+  push:
+  pull_request:
+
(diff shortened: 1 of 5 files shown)


leoromanovsky (approved):
Looks like dropping down is a good choice; other vendors (https://github.com/launchdarkly/rust-server-sdk/blob/main/launchdarkly-server-sdk/Cargo.toml#L7) are at 1.7.0.

I am going to squash in case a revert is necessary.

PR #7, opened 2024-05-15
feat: add typed versions of get_assignment()

rasendubi:


rasendubi:
I included https://github.com/Eppo-exp/rust-sdk/pull/7/commits/f754132fbdc0cac6b0ce780eead6bb7449e82873 to minimize merge conflict with #6.

src/client.rs
```diff
@@ -103,23 +117,293 @@ impl<'a> Client<'a> {
...
+    /// You should call [`Client::start_poller_thread`] before any call to
+    /// `get_boolean_assignment()`. Otherwise, the client will always return `None`.
+    ///
+    /// It is recommended to wait for the Eppo configuration to get fetched with
+    /// [`PollerThread::wait_for_configuration()`].
+    ///
+    /// # Errors
+    ///
+    /// Returns an error in the following cases:
+    /// - [`Error::FlagNotFound`] if the requested flag configuration was not found.
```
leoromanovsky, line 287: This could happen if someone adds the `get_*_assignment` function to their code but hasn't yet configured a flag in eppo.cloud. This is a normal flow and from the perspective of the client, should not throw an error.

What I want to happen, from the client's perspective, is the `default_value` be returned. However the `FlagNotFound` can be returned in some way to inform users of the reason for the value.

src/client.rs
```diff
@@ -103,23 +117,293 @@ impl<'a> Client<'a> {
...
+    ///     ].into_iter().collect())
+    ///     .unwrap_or_default()
+    ///     .unwrap_or("default_value".to_owned());
+    /// # }
+    /// ```
+    pub fn get_string_assignment(
+        &self,
+        flag_key: &str,
+        subject_key: &str,
+        subject_attributes: &SubjectAttributes,
```
leoromanovsky, line 158: Please add a `default_value: String` parameter, or whatever the type is returned, to all public `get_*_assignment` methods. This will match our APIs in other SDKs: https://github.com/Eppo-exp/js-client-sdk-common/blob/main/src/client/eppo-client.ts#L45-L78

leoromanovsky (approved):

PR #11, opened 2024-07-03
feat: bandits

rasendubi:
WIP: missing testing on Ruby side


rasendubi:
Rust tests will fail depending on https://github.com/Eppo-exp/sdk-test-data/pull/44

ruby-sdk/ext/eppo_rb/Cargo.toml
```diff
@@ -10,7 +10,7 @@ rust-version = "1.65.0"
 crate-type = ["cdylib"]
 
 [dependencies]
-env_logger = { version = "0.11.3", features = ["unstable-kv"] }
+env_logger = { version = "0.11.3", default-features = false, features = ["unstable-kv"] }
```
rasendubi, line 13: fix for MSRV

eppo_core/src/bandits/eval.rs
```diff
@@ -0,0 +1,421 @@
...
+                flag_key,
+                subject_key,
+                &subject_attributes.to_generic_attributes(),
+                Some(VariationType::String),
+            )
+            .unwrap_or_default()
+            .unwrap_or_else(|| Assignment {
+                value: AssignmentValue::String(default_variation.to_owned()),
+                event: None,
+            });
```
rasendubi, line 78: One deviation from other SDKs: because of how `eppo_core` is structured, it is possible to evaluate assignments without triggering assignment logging. I therefore removed an early return in case it's a bandit flag but actions are missing and evaluate the assignment immediately.
- In case assignment resolves to non-bandit variation, we return the variation and log the assignment.
- If it's a bandit variation and actions are missing, bandit evaluation will fail and we'll return default variation and skip assignment logging.
- Otherwise, we return variation+action + log the assignment and bandit.

aarsilv: I think you could mirror the other SDKs if you stored top-level which flags have which bandits. Then you know if a flag involves bandits or not from the get go and can check if assignments is empty.

I do agree it's kind of a pain though.

rasendubi: Yeah, I store that mapping and I can easily mirror the behavior, but I thought that this behavior is more desirable from the user perspective (they receive correct variation instead of the default one).

PR #19, opened 2024-07-24
FF-2853 fix: rename eppo_rb to eppo_client

rasendubi:
For some reason, ExtensionTask ignored lib_dir during gem installation, so extension was installed to lib/eppo_rb and gem failed to function.

Renaming import does not help because it breaks the build in local development.

Rename eppo_rb to eppo_client, so extension name matches the installation directory. This is also what was recommended by rust gem template.


greghuels (approved):

leoromanovsky (approved):
tested locally ✔️

PR #20, opened 2024-07-24
FF-2857 chore: add publish to rubygems

leoromanovsky:
* Adds ability to publish to ruby gems
* Continues with publish if a dependent version already exists
* Added `RUBYGEMS_API_KEY` environment variable to github repo with permission to publish to https://rubygems.org/gems/eppo-server-sdk

**Testing**

Ran the steps manually: https://rubygems.org/gems/eppo-server-sdk

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -36,3 +36,29 @@ jobs:
         run: cargo publish -p eppo
         env:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CRATES_IO_TOKEN }}
+
+  publish_ruby:
+    name: Publish to RubyGems
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '3.0'
+      - name: Install dependencies
+        run: bundle install
+        working-directory: ruby-sdk
+      - name: Build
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -36,3 +36,29 @@ jobs:
         run: cargo publish -p eppo
         env:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CRATES_IO_TOKEN }}
+
+  publish_ruby:
```
leoromanovsky, line 40: Inspired by https://github.com/Eppo-exp/ruby-sdk/blob/main/.github/workflows/publish.yml and updated versions

rasendubi:
The approach here has two issues:

1. It tries to publish eppo_core and ruby sdk in parallel which may fail if
ruby sdk depends on newer version of eppo_core

2. Because ruby extension requires us updating the lockfile after eppo_core
is published, they can't be run in the same workflow (unless it updates and
commits the lockfile before publishing ruby sdk)

We can work on a more seamless multi-package publishing in the future but
for now I would suggest splitting the workflow per-package (e.g. by tag
name: ***@***.***, ***@***.***, etc)

rasendubi:
Email reply garbled my comment. I meant eppo_core-1.0.0, ruby-sdk-3.0.0,
etc.

rasendubi:
I have updated the PR with filtering by tag name.

greghuels (approved):

PR #13, opened 2024-07-17
FF-2678 flag evaluation details

rasendubi:
https://linear.app/eppo/issue/FF-2678/evaluation-reasons-in-rust

Examples:
<details>
<summary>Example 1</summary>

```json
{
  "variation": true,
  "action": null,
  "evaluationDetails": {
    "flagKey": "a-boolean-flag",
    "subjectKey": "test-subject",
    "subjectAttributes": {
      "name": "oleksii"
    },
    "timestamp": "2024-07-22T18:02:20.122461Z",
    "configFetchedAt": "2024-07-22T18:02:20.119761Z",
    "configPublishedAt": "2024-07-22T17:41:28.164Z",
    "environmentName": "eppo.cloud",
    "flagEvaluationCode": "MATCH",
    "variationKey": "true",
    "variationValue": true,
    "banditKey": null,
    "banditAction": null,
    "allocations": [
      {
        "key": "allocation-4233",
        "orderPosition": 1,
        "allocationEvaluationCode": "FAILING_RULE",
        "evaluatedRules": [
          {
            "matched": false,
            "conditions": [
              {
                "condition": {
                  "operator": "MATCHES",
                  "attribute": "email",
                  "value": ".*@example\\.com"
                },
                "attributeValue": null,
                "matched": false
              }
            ]
          }
        ],
        "evaluatedSplits": []
      },
      {
        "key": "allocation-4246",
        "orderPosition": 2,
        "allocationEvaluationCode": "MATCHED",
        "evaluatedRules": [
          {
            "matched": true,
            "conditions": [
              {
                "condition": {
                  "operator": "ONE_OF",
                  "attribute": "name",
                  "value": [
                    "leo",
                    "oleksii",
                    "felipe"
                  ]
                },
                "attributeValue": "oleksii",
                "matched": true
              }
            ]
          }
        ],
        "evaluatedSplits": [
          {
            "variationKey": "true",
            "matched": true,
            "shards": [
              {
                "matched": true,
                "shard": {
                  "salt": "a-boolean-flag-4246-split",
                  "ranges": [
                    {
                      "start": 0,
                      "end": 10000
                    }
                  ]
                },
                "shardValue": 4610
              }
            ]
          }
        ]
      },
      {
        "key": "allocation-4446",
        "orderPosition": 3,
        "allocationEvaluationCode": "UNEVALUATED",
        "evaluatedRules": [],
        "evaluatedSplits": []
      },
      {
        "key": "allocation-4447",
        "orderPosition": 4,
[...]

PR #22, opened 2024-07-29
chore: add sdk-test-data as a submodule

rasendubi:
I'm somewhat tired of `sdk-test-data` updates breaking this SDK unprompted. I also want the ability to apply test-data fixes in branches (e.g., #13 is currently blocked by https://github.com/Eppo-exp/sdk-test-data/pull/49).

Plan:
- add `sdk-test-data` as a submodule so that we can update it when we're ready (instead of updates being pushed on us unexpectedly). (this PR)
- make [Eppo-exp/sdk-test-data](https://github.com/Eppo-exp/sdk-test-data) open a pull request against [Eppo-exp/rust-sdk](https://github.com/Eppo-exp/rust-sdk) when test data change. This will allow our CI validate tests and gives us a chance to adapt our code to new tests before we accept the PR. (Will open a PR on sdk-test-data after this one is merged.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,9 +18,10 @@ jobs:
         toolchain:
           - stable
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
       - run: npm ci
-      - run: make test-data
       - run: rustup update ${{ matrix.toolchain }} && rustup default ${{ matrix.toolchain }}
       - run: cargo build --verbose
       - run: cargo test --verbose
@@ -37,6 +38,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: true
 
       - name: Set up Ruby & Rust
         uses: oxidize-rb/actions/setup-ruby-and-rust@v1
@@ -47,7 +50,6 @@ jobs:
           rubygems: '3.5.11'
 
       - run: npm ci
-      - run: make test-data
 
       - name: Check Cargo.lock
         # Ensure that Cargo.lock matches Cargo.toml
--- a/.github/workflows/msrv.yml
+++ b/.github/workflows/msrv.yml
@@ -19,7 +19,9 @@ jobs:
           - stable
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
       - uses: actions/cache@v2
         with:
           path: |
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -29,12 +29,13 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ startsWith(github.ref_name, 'eppo_core@') || startsWith(github.ref_name, 'rust-sdk@') }}
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
       - uses: actions/setup-node@v3
         with:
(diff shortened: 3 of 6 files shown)


aarsilv:
> I'm somewhat tired of sdk-test-data updates breaking this SDK unprompted

Agree this is annoying, however given the sdk-test-data is the source of truth regarding "correct" operation, it seems like its basically an annoying way of notifying that changes have been decided. However, one way or another, we'd want that notification.

> make [Eppo-exp/sdk-test-data](https://github.com/Eppo-exp/sdk-test-data) open a pull request against [Eppo-exp/rust-sdk](https://github.com/Eppo-exp/rust-sdk) when test data change

I'm not sure this is the plan we want. Would this then make the rust SDK different than others?

If we do want to tie the SDK to snapshots of our test data (which I don't think we do), I wonder if just specifying a version hash in the checkout step would be easier

rasendubi:
> Agree this is annoying, however given the sdk-test-data is the source of truth regarding "correct" operation, it seems like its basically an annoying way of notifying that changes have been decided. However, one way or another, we'd want that notification.

Yes. I want that notification, too—just want to make it less "annoying."

I wouldn't call annoyance as the main reason though—my main concern is that it's breaking too much. It breaks `main` branch, all open pull requests, and all previous commits _retroactively._

Besides hindering the development flow (having to take a look why CI on my PR or local tests suddenly go bonkers), it also breaks some other less frequently used tools: `git bisect` doesn't work, re-triggering a CI on old commit will break it, it might potentially break a release process if timing is bad, etc. (Versioning tests along the source code is always a good idea.)

Besides that, this approach is bad at notifications, too—if an SDK doesn't have new PRs open, there's no notification.

Opening a PR is a better way to notify about changes elsewhere (e.g., dependabot is a widely known example).

> Would this then make the rust SDK different than others?

Hm, yes, Rust (and Ruby 😼) would differ in a small way.

If we have a good experience here though, I believe we would want to replicate that behavior to other SDKs.

> If we do want to tie the SDK to snapshots of our test data (which I don't think we do), I wonder if just specifying a version hash in the checkout step would be easier

Not sure I follow. A git submodule _is_ just specifying a version hash in the checkout step. What do you mean?

Just to clarify my proposal: when `sdk-test-data`'s `main` branch updates, that would open a PR for this repo, updating a single hash of submodule to latest value (i.e., we get the notification regardless of repo activity). If the CI passes, we merge it (or maybe have it merged automatically). If CI fails, we can fix the SDK before merging. We got the notification, `main` remains green, PRs are not blocked, everyone's happy 🎉

aarsilv:
I'm probably misunderstanding how submodules work. If basically a change in sdk-test-data results in an auto-created PR for this repository to update to use the latest test data, and leaves all other repositories unaffected, then this does sound good.

It sounds conceptually similar to my suggestion, which is to have make-test-data check out a specific version of sdk-test-data, so a PR would be needed to get more recent data. But no automatic PR as in your proposal, which I think we want.

rasendubi:
Submodule is basically a pointer to another repository commit (url + commit hash) that git can pull for you. Seems to be exactly your proposal—just using native git tool.

Because it stores a specific hash, we need some way to update it when upstream `sdk-test-data` changes. Submodule doesn't do anything by itself, so my proposal is for `sdk-test-data` CI to open a PR against this repo with new hash.

aarsilv:
Ok I see, so it would be the combination of the sub module plus an action in the test data repository that automatically opens a PR? Sounds good to try this out here and if we like it, employ it in other repositories.

aarsilv (approved):
Thanks for answering my questions!

PR #23, opened 2024-08-09
feat(eppo_core): allow configuring poller thread

rasendubi:
This is a minor change for eppo_core. It does not affect any existing clients.

### DIFF
--- a/eppo_core/src/poller_thread.rs
+++ b/eppo_core/src/poller_thread.rs
@@ -11,6 +11,54 @@ use crate::configuration_fetcher::ConfigurationFetcher;
 use crate::configuration_store::ConfigurationStore;
 use crate::{Error, Result};
 
+/// Configuration for [`PollerThread`].
+// Not implementing `Copy` as we may add non-copyable fields in the future.
+#[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


eppo_core/src/poller_thread.rs
```diff
@@ -79,7 +143,7 @@ impl PollerThread {
                             }
                         };
 
-                        let timeout = jitter(POLL_INTERVAL, POLL_JITTER);
+                        let timeout = config.interval + jitter(config.jitter);
```
schmit, line 146: Nit: we should subtract the jitter to make sure we always poll within the interval as set up (the interval is an upper bound instead of a lower bound)

schmit: Note: will have to make sure that jitter < interval ( / 2?)

schmit (approved):

eppo_core/src/poller_thread.rs
rasendubi: Good catch! I somehow missed that jitter should be subtracted during initial implementation

I have fixed that and added two tests: to check that jitter is subtractive + that underflow (`interval - jitter < 0`) is handled correctly

PR #25, opened 2024-08-14
chore: bump versions before release

rasendubi:


.github/workflows/ci.yml
```diff
@@ -51,11 +51,6 @@ jobs:
 
       - run: npm ci
 
-      - name: Check Cargo.lock
```
rasendubi, line 54: This check failed on PR because `ruby-sdk` now relies on yet-unpublished `eppo_core@2.0.0`.

I have moved this check to `publish` workflow because we only need the lock file for published version

greghuels (approved):

PR #31, opened 2024-09-13
FF-3204 perf: compile regex during parsing

rasendubi:
After benchmarking Rust core and Python SDK, I noticed that rust core performs quite badly on regex conditions. Immediate reaction: regex compilation is kinda slow and we do it on every evaluation call.

This PR moves regex compilation from evaluation phase to parsing, so it happens only once and in background (doesn't slow down evaluation calls).

## Description of changes
- This PR adds `ConditionWire` struct, which is the structure we receive from the server (what is transferred "on the wire").
- `Condition` is a new type and optimized for evaluation. There are function to convert back and forth between the two formats. In additional to parsing regex, we also prepare other conditions:
  - For comparison operator, we parse version string beforehand
  - Check that condition value matches the condition operator (e.g., `IS_NULL` expects a boolean condition value)
- Evaluation details is adjusted to store `ConditionWire`, so the change is not visible to users

## Results

Comparing against pre-optimization, regex evaluation is ~150x faster now (from ~45us down to 300ns):

```
regex-flag/get_assignment
                        time:   [295.74 ns 296.20 ns 296.71 ns]
                        thrpt:  [3.3703 Melem/s 3.3761 Melem/s 3.3813 Melem/s]
                 change:
                        time:   [-99.355% -99.354% -99.352%] (p = 0.00 < 0.05)
                        thrpt:  [+15338% +15369% +15403%]
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  7 (7.00%) high mild

regex-flag/get_assignment_details
                        time:   [1.8295 µs 1.8331 µs 1.8369 µs]
                        thrpt:  [544.40 Kelem/s 545.53 Kelem/s 546.60 Kelem/s]
                 change:
                        time:   [-96.137% -96.112% -96.093%] (p = 0.00 < 0.05)
                        thrpt:  [+2459.8% +2472.3% +2488.7%]
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe
```

For python SDK, regex benchmark used to be 5x slower than native python sdk. Now it is 14x faster:
```
+----------------------------------+---------+-----------------------+-----------------------+
| Benchmark                        | 3.7.0   | 4.0.0-old             | 4.0.0                 |
+==================================+=========+=======================+=======================+
| Evaluation (sdk-test-data)       | 14.3 us | 7.10 us: 2.01x faster | 3.75 us: 3.81x faster |
+----------------------------------+---------+-----------------------+-----------------------+
| Evaluation (new-user-onboarding) | 9.27 us | 2.37 us: 3.92x faster | 361 ns: 25.67x faster |
+----------------------------------+---------+-----------------------+-----------------------+
| Evaluation (numeric-one-of)      | 9.77 us | 774 ns: 12.62x faster | 771 ns: 12.67x faster |
+----------------------------------+---------+-----------------------+-----------------------+
| Evaluation (regex-flag)          | 8.43 us | 46.1 us: 5.47x slower | 567 ns: 14.86x faster |
+----------------------------------+---------+-----------------------+-----------------------+

```

### DIFF
--- a/eppo_core/benches/evaluation_details.rs
+++ b/eppo_core/benches/evaluation_details.rs
@@ -128,6 +128,35 @@ fn criterion_benchmark(c: &mut Criterion) {
         });
         group.finish();
     }
+
+    {
+        let mut group = c.benchmark_group("regex-flag");
+        group.throughput(Throughput::Elements(1));
+        let attributes = [("email".into(), "test@gmail.com".into())].into();
+        group.bench_function("get_assignment", |b| {
+            b.iter(|| {
+                get_assignment(
+                    black_box(Some(&configuration)),
+                    black_box("regex-flag"),
+                    black_box("subject1"),
+                    black_box(&attributes),
+                    black_box(None),
+                )
+            })
+        });
+        group.bench_function("get_assignment_details", |b| {
+            b.iter(|| {
+                get_assignment_details(
+                    black_box(Some(&configuration)),
+                    black_box("regex-flag"),
+                    black_box("subject1"),
+                    black_box(&attributes),
+                    black_box(None),
+                )
+            })
+        });
+        group.finish();
+    }
 }
 
 criterion_group!(
--- a/eppo_core/src/eval/eval_rules.rs
+++ b/eppo_core/src/eval/eval_rules.rs
@@ -1,8 +1,9 @@
-use regex::Regex;
+use std::borrow::Cow;
+
 use semver::Version;
 
 use crate::{
-    ufc::{Condition, ConditionOperator, ConditionValue, Rule, Value},
+    ufc::{Comparand, ComparisonOperator, Condition, ConditionCheck, Rule, TryParse},
     AttributeValue, Attributes,
 };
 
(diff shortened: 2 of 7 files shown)


eppo_core/benches/evaluation_details.rs
```diff
@@ -128,6 +128,35 @@ fn criterion_benchmark(c: &mut Criterion) {
         });
         group.finish();
     }
+
+    {
+        let mut group = c.benchmark_group("regex-flag");
```
aarsilv, line 133: 🔥

eppo_core/src/eval/eval_rules.rs
```diff
@@ -14,128 +15,97 @@ impl Rule {
...
         attributes: &Attributes,
     ) -> bool {
-        self.conditions
-            .iter()
-            .all(|condition| condition.eval(visitor, attributes))
+        self.conditions.iter().all(|condition| match condition {
+            TryParse::Parsed(condition) => condition.eval(visitor, attributes),
+            TryParse::ParseFailed(raw_condition) => {
+                visitor.on_condition_skip(raw_condition);
+                false
```
aarsilv, line 22: ok cool so by returning false we won't match when a condition is skipped due to failing to parse 👌

eppo_core/src/eval/eval_rules.rs
```diff
@@ -14,128 +15,97 @@ impl Rule {
...
+                            "true"
+                        } else {
+                            "false"
+                        }
+                    }
                     _ => return None,
                 };
-                let values = match condition_value {
-                    ConditionValue::Multiple(v) => v,
+                regex.is_match(s) == *expected_match
```
aarsilv, line 87: 🏎️

eppo_core/src/ufc/models.rs
```diff
@@ -45,8 +49,16 @@ pub enum TryParse<T> {
     /// Successfully parsed.
     Parsed(T),
     /// Parsing failed.
+    ///
+    /// This holds the generic JSON value, so even if parsing originally failed, we could serialize
```
aarsilv, line 53: 👍

eppo_core/src/ufc/models.rs
```diff
@@ -194,24 +206,242 @@ fn default_do_log() -> bool {
...
+pub struct ConditionWire {
+    pub attribute: Box<str>,
     pub operator: ConditionOperator,
-    pub attribute: String,
     pub value: ConditionValue,
 }
 
+impl From<Condition> for ConditionWire {
+    fn from(condition: Condition) -> Self {
+        let (operator, value) = match condition.check {
```
aarsilv, line 293: cleaner having this stuff pulled out 🧹

eppo_core/src/ufc/models.rs
```diff
@@ -194,24 +206,242 @@ fn default_do_log() -> bool {
...
+                        log::warn!(
+                            "failed to parse condition: {:?} condition with non-string condition value",
+                            condition.operator
+                        );
+                        return Err(Error::EvaluationError(
+                            EvaluationError::UnexpectedConfigurationParseError,
+                        ));
+                    }
+                };
+                let regex = match Regex::new(&regex_string) {
```
aarsilv, line 359: 💪

aarsilv (approved):
Looking good to me!

leoromanovsky (approved):
great insight and thanks for keeping the API the same while you made these changes.

PR #32, opened 2024-09-18
refactor(all),fix(rust): introduce Evaluator and handle SdkMetadata

rasendubi:
Previously, eppo_core did not set metadata on events correctly (sdk name and version were missing). Python and rust accounted for this separately, and Rust didn't handle it at all.

This PR makes evaluation function take SdkMetadata as a parameter, so it's now impossible to miss.

`now` is not also a parameter, so all evaluation functions are now pure functions without side effects, so we can add time-based tests in the future.

This PR also introduces Evaluator to simplify calling into evaluator functions for upstream SDKs.

### DIFF
--- a/eppo_core/src/eval/evaluator.rs
+++ b/eppo_core/src/eval/evaluator.rs
@@ -1,0 +1,120 @@
+use std::{collections::HashMap, sync::Arc};
+
+use chrono::Utc;
+
+use crate::{
+    configuration_store::ConfigurationStore,
+    events::AssignmentEvent,
+    ufc::{Assignment, AssignmentValue, VariationType},
+    Attributes, Configuration, ContextAttributes, EvaluationError, SdkMetadata,
+};
+
+use super::{
+    eval_details::{EvaluationDetails, EvaluationResultWithDetails},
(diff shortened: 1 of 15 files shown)


eppo_core/src/eval/evaluator.rs
```diff
@@ -0,0 +1,120 @@
...
+    ) -> (BanditResult, EvaluationDetails) {
+        let configuration = self.get_configuration();
+        get_bandit_action_details(
+            configuration.as_ref().map(|it| it.as_ref()),
+            flag_key,
+            subject_key,
+            subject_attributes,
+            actions,
+            default_variation,
+            Utc::now(),
```
leoromanovsky, line 112: Is this in the same date format as the assignment logger?

leoromanovsky (approved):
Thank you for improving the core

eppo_core/src/eval/evaluator.rs
rasendubi: `Utc::now()` returns `DateTime<Utc>` which is a raw date-time value. It gets formatted later in another place

PR #33, opened 2024-09-18
doc(core): expand overview documentation

rasendubi:


leoromanovsky (approved):
Fantastic - what additional benefit does the documentation in a rust file gain us over markdown?

rasendubi:
> what additional benefit does the documentation in a rust file gain us over markdown?

It's easier to reference types and functions from a documentation comment, so I can link to what I talk about. It will also get published to https://docs.rs/eppo_core/ once crate is released.

PR #38, opened 2024-09-27
FF-3205 perf: minimize copying and a bunch of other optimizations

rasendubi:
This PR minimizes copying that happens inside SDK. This is mostly done by using `Arc` (atomic reference counted pointer).

For strings, I used `Arc<str>` initially. However, I looked through a ton of crates and found one that is optimized for our use case (cheap cloning) — [faststr](https://docs.rs/faststr). In addition to allowing `Arc<str>`, it tries to minimize allocation when converting to/from string type and supports small string optimization (embedding strings up to 24 bytes inside the value itself, without memory allocation). I have benchmarked it and it gives even better results.

Another big realization is that most of `AssignmentEvent` object is static for the same split, so we can pre-compute events beforehand, so we don't move tons of strings around. After we select a split, we can directly get a pair of (assignment value, event).

I also ported JSON precomputing that we implemented in Go SDK (which gives a 5x boost on our json assignment microbenchmark).

As part of implementing pre-computing, I also realized that we can pre-hash salt values, so this gives a small boost for sharding.

The combined effect is a 2.5–4x boost (even on non-logging assignments). Assignment details are having a modest 20% boost.

Python in particular is enjoying these changes (I assume due to decreased allocation on Python/Rust boundary):
```
+----------------------------------+--------------+-----------------------+
| Benchmark                        | 4.0.0-before | 4.0.0-after           |
+==================================+==============+=======================+
| Evaluation                       | 11.5 us      | 2.85 us: 4.05x faster |
+----------------------------------+--------------+-----------------------+
| Evaluation (new-user-onboarding) | 2.96 us      | 261 ns: 11.31x faster |
+----------------------------------+--------------+-----------------------+
| Evaluation (numeric-one-of)      | 5.87 us      | 735 ns: 7.99x faster  |
+----------------------------------+--------------+-----------------------+
| Evaluation (regex-flag)          | 5.09 us      | 504 ns: 10.10x faster |
+----------------------------------+--------------+-----------------------+
| Geometric mean                   | (ref)        | 7.79x faster          |
+----------------------------------+--------------+-----------------------+
```

### DIFF
--- a/eppo_core/benches/evaluation_details.rs
+++ b/eppo_core/benches/evaluation_details.rs
@@ -1,5 +1,5 @@
-use std::collections::HashMap;
 use std::fs::File;
+use std::{collections::HashMap, sync::Arc};
 
 use chrono::Utc;
 use criterion::{black_box, criterion_group, criterion_main, Criterion, Throughput};
@@ -22,13 +22,13 @@ fn criterion_benchmark(c: &mut Criterion) {
     {
         let mut group = c.benchmark_group("new-user-onboarding");
         group.throughput(Throughput::Elements(1));
-        let attributes = HashMap::new();
+        let attributes = Arc::new(HashMap::new());
         group.bench_function("get_assignment", |b| {
             b.iter(|| {
                 get_assignment(
                     black_box(Some(&configuration)),
                     black_box("new-user-onboarding"),
-                    black_box("subject1"),
+                    black_box(&"subject1".into()),
                     black_box(&attributes),
                     black_box(None),
                     black_box(now),
@@ -41,7 +41,7 @@ fn criterion_benchmark(c: &mut Criterion) {
                 get_assignment_details(
                     black_box(Some(&configuration)),
                     black_box("new-user-onboarding"),
-                    black_box("subject1"),
+                    black_box(&"subject1".into()),
                     black_box(&attributes),
                     black_box(None),
                     black_box(now),
@@ -55,13 +55,13 @@ fn criterion_benchmark(c: &mut Criterion) {
     {
         let mut group = c.benchmark_group("rollout");
         group.throughput(Throughput::Elements(1));
-        let attributes = [("country".to_owned(), "US".into())].into();
+        let attributes = Arc::new([("country".to_owned(), "US".into())].into());
         group.bench_function("get_assignment", |b| {
             b.iter(|| {
                 get_assignment(
                     black_box(Some(&configuration)),
                     black_box("new-user-onboarding"),
-                    black_box("subject1"),
+                    black_box(&"subject1".into()),
                     black_box(&attributes),
                     black_box(None),
                     black_box(now),
@@ -74,7 +74,7 @@ fn criterion_benchmark(c: &mut Criterion) {
                 get_assignment_details(
                     black_box(Some(&configuration)),
                     black_box("new-user-onboarding"),
-                    black_box("subject1"),
+                    black_box(&"subject1".into()),
                     black_box(&attributes),
                     black_box(None),
                     black_box(now),
@@ -88,13 +88,13 @@ fn criterion_benchmark(c: &mut Criterion) {
     {
         let mut group = c.benchmark_group("json-config-flag");
         group.throughput(Throughput::Elements(1));
-        let attributes = [].into();
+        let attributes = Arc::new([].into());
         group.bench_function("get_assignment", |b| {
             b.iter(|| {
                 get_assignment(
                     black_box(Some(&configuration)),
                     black_box("json-config-flag"),
-                    black_box("subject1"),
+                    black_box(&"subject1".into()),
                     black_box(&attributes),
                     black_box(None),
                     black_box(now),
@@ -107,7 +107,7 @@ fn criterion_benchmark(c: &mut Criterion) {
                 get_assignment_details(
                     black_box(Some(&configuration)),
                     black_box("json-config-flag"),
-                    black_box("subject1"),
+                    black_box(&"subject1".into()),
                     black_box(&attributes),
                     black_box(None),
                     black_box(now),
@@ -121,13 +121,13 @@ fn criterion_benchmark(c: &mut Criterion) {
     {
         let mut group = c.benchmark_group("numeric-one-of");
         group.throughput(Throughput::Elements(1));
-        let attributes = [("number".to_owned(), 2.0.into())].into();
+        let attributes = Arc::new([("number".to_owned(), 2.0.into())].into());
         group.bench_function("get_assignment", |b| {
             b.iter(|| {
                 get_assignment(
                     black_box(Some(&configuration)),
                     black_box("numeric-one-of"),
-                    black_box("subject1"),
+                    black_box(&"subject1".into()),
                     black_box(&attributes),
                     black_box(None),
                     black_box(now),
@@ -140,7 +140,7 @@ fn criterion_benchmark(c: &mut Criterion) {
                 get_assignment_details(
                     black_box(Some(&configuration)),
                     black_box("numeric-one-of"),
-                    black_box("subject1"),
+                    black_box(&"subject1".into()),
                     black_box(&attributes),
                     black_box(None),
                     black_box(now),
@@ -154,13 +154,13 @@ fn criterion_benchmark(c: &mut Criterion) {
     {
         let mut group = c.benchmark_group("regex-flag");
         group.throughput(Throughput::Elements(1));
-        let attributes = [("email".into(), "test@gmail.com".into())].into();
+        let attributes = Arc::new([("email".into(), "test@gmail.com".into())].into());
         group.bench_function("get_assignment", |b| {
             b.iter(|| {
                 get_assignment(
                     black_box(Some(&configuration)),
                     black_box("regex-flag"),
-                    black_box("subject1"),
+                    black_box(&"subject1".into()),
                     black_box(&attributes),
                     black_box(None),
                     black_box(now),
@@ -173,7 +173,7 @@ fn criterion_benchmark(c: &mut Criterion) {
                 get_assignment_details(
                     black_box(Some(&configuration)),
                     black_box("regex-flag"),
-                    black_box("subject1"),
+                    black_box(&"subject1".into()),
                     black_box(&attributes),
                     black_box(None),
                     black_box(now),
--- a/eppo_core/benches/evaluation_details.rs
+++ b/eppo_core/benches/evaluation_details.rs
@@ -1,23 +1,25 @@
-use std::fs::File;
 use std::{collections::HashMap, sync::Arc};
 
 use chrono::Utc;
 use criterion::{black_box, criterion_group, criterion_main, Criterion, Throughput};
 
+use eppo_core::ufc::UniversalFlagConfig;
 use eppo_core::{
     eval::{get_assignment, get_assignment_details},
     Configuration, SdkMetadata,
 };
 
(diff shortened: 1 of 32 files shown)


eppo_core/src/attributes.rs
```diff
@@ -1,8 +1,48 @@
...
+use crate::Str;
+
+/// `Subject` is a bundle of subject attributes and a key.
+#[derive(Debug)]
+pub(crate) struct Subject {
+    /// Subject key encoded as attribute value. Known to be `AttributeValue::String`. This is
+    /// done to allow returning subject key as an attribute when rule references "id".
+    key: AttributeValue,
+    attributes: Arc<Attributes>,
+}
```
rasendubi, line 15: The `Subject` is here to avoid copying subject attributes to add the implicit "id" attribute.

eppo_core/src/configuration.rs
```diff
@@ -1,25 +1,21 @@
...
     /// Timestamp when configuration was fetched by the SDK.
     pub fetched_at: DateTime<Utc>,
     /// Flags configuration.
     pub flags: UniversalFlagConfig,
     /// Bandits configuration.
     pub bandits: Option<BanditResponse>,
-    /// Mapping from flag key to flag variation value to bandit variation. Cached from
-    /// `self.flags.bandits`.
-    pub flag_to_bandit_associations:
-        HashMap</* flag_key: */ String, HashMap</* variation_key: */ String, BanditVariation>>,
```
rasendubi, line 22: This has been moved to pre-compiled `UniversalFlagConfig` directly.

eppo_core/src/eval/eval_assignment.rs
```diff
@@ -214,84 +205,32 @@ impl Flag {
...
-                (
-                    "eppoCoreVersion".to_owned(),
-                    env!("CARGO_PKG_VERSION").to_owned(),
-                ),
-            ]
-            .into(),
-            extra_logging: split.extra_logging.clone(),
-            evaluation_details: None,
-        });
+        let (value, event_base) = split.result.clone()?;
```
leoromanovsky, line 223: 📓 fetch pre-computed assignment event

eppo_core/src/ufc/compiled_flag_config.rs
```diff
@@ -0,0 +1,265 @@
...
+) -> Split {
+    let shards = split
+        .shards
+        .into_iter()
+        .filter_map(|shard|
+                            // `compile_shard` may return `None` for shards that are
+                            // "insignificant", meaning that they *always* match, so they don't even
+                            // need to be checked. We filter out such shards here with
+                            // `.filter_map()`.
+                            compile_shard(shard, total_shards))
```
leoromanovsky, line 221: 📓 filtering out insignificant shards

leoromanovsky (commented):
Took a read; left a few annotations for another pass with Sven tomorrow.

eppo_core/src/sharder.rs
```diff
@@ -1,11 +1,47 @@
...
+}
+
+impl std::fmt::Debug for PreSaltedSharder {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("PreSaltedSharder")
+            .field("ctx", &"...")
+            .field("total_shards", &self.total_shards)
+            .finish()
+    }
+}
```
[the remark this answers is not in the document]
rasendubi: A bit more explanation.

md5 hash works by keeping an internal state (aka context, just a bunch of bytes) and updating it with every input bytes. This process is deterministic, so consuming same bytes always produce the same state. After input is done, the state is "finalized" and the final hash value is derived from it.

What happens here is that we consume the salt with "-" and save the md5 context (`new()` in pre-processing step). When we get subject key in evaluation step, we just copy saved (pre-salted) context, consume subject key, and finalize the hash (`shard()`).

eppo_core/src/ufc/compiled_flag_config.rs
rasendubi: More explanation.

A split is selected when the subject matches all shards in the split. i.e., to match a split, the subject must satisfy *all* shards; if there're no shards in the split, that's an automatic match.

So we have the following formula:
```
matches_split = true && matches_shard1 && matches_shard2 && ... && matches_shardN
```

A shard that contains a $[0..total\\_shards)$ range always evaluates to `true` because the shard value will always fall in that range ($\forall hash \in \mathbb{N}: 0 \le hash\ \\%\ total\\_shards < total\\_shards$), and that range will dominate the shard (because shard matches if any range matches (OR), and $\forall x: x \lor true = true$).

So we can replace always-matching shard with a `true` in the original equation:

```
matches_split = true && matches_shard1 && true && ... && matches_shardN
```

And because we know that `true` is an identity for `&&` ($\forall x: x \land true = x$), we can drop `true`s from the formula. So we can drop that shard altogether:

```
matches_split = matches_shard1 && ... && matches_shardN
```

eppo_core/src/eval/eval_assignment.rs
rasendubi: The `event_base` is precomputed and contains the following fields:
```rust
struct AssignmentEventBase {
    feature_flag: Str, // <- flag key
    allocation: Str, // <- allocation key
    experiment: String, // <- {flag}-{allocation}
    variation: Str, // variation key from split
    meta_data: EventMetaData, // constant {sdk_name,sdk_version,core_version}
    extra_logging: HashMap<String, String>, // extra logging from split
}
```

None of these values are dependent on subject or specific evaluation call, and the split determines all the fields here. Assignment value is also dictated by the selected split.

More abstractly, we pre-compute a map of {split -> evaluation result} (technically, a decision tree), and evaluation is just a matter of selecting the split and returning the associated result.

eppo_core/benches/evaluation_details.rs
```diff
@@ -1,38 +1,39 @@
...
 };
 
 fn criterion_benchmark(c: &mut Criterion) {
-    let flags =
-        serde_json::from_reader(File::open("../sdk-test-data/ufc/flags-v1.json").unwrap()).unwrap();
+    let flags = UniversalFlagConfig::from_json(
+        SdkMetadata {
+            name: "test",
+            version: "0.1.0",
+        },
```
schmit, line 17: Are these values placeholders?

eppo_core/src/attributes.rs
```diff
@@ -1,8 +1,48 @@
...
+
+    pub fn get_attribute(&self, name: &str) -> Option<&AttributeValue> {
+        let value = self.attributes.get(name);
+        if value.is_some() {
+            return value;
+        }
+
+        if name == "id" {
+            return Some(&self.key);
+        }
```
schmit, line 40: I imagine the ordering is such that when someone explicitly sets `"id"`, that takes precedence, which is great! But I think it might be nice to add that as a comment for our future selves

eppo_core/src/attributes.rs
```diff
@@ -67,19 +108,19 @@ mod pyo3_impl {
     impl<'py> FromPyObject<'py> for AttributeValue {
         fn extract_bound(value: &Bound<'py, PyAny>) -> PyResult<AttributeValue> {
             if let Ok(s) = value.downcast::<PyString>() {
-                return Ok(AttributeValue::String(s.extract()?));
+                return Ok(AttributeValue::String(s.to_cow()?.into()));
```
schmit, line 111: Trying to understand the inner workings of Rust here, hope you are willing to entertain me. According to ChatGPT:

```
If the string is always going to be used as owned data, you could bypass the Cow<str> entirely and directly convert it into a String. However, the current approach is more flexible and performance-conscious, as it handles both borrowed and owned cases efficiently.
```

Do we need the flexibility to handle both? I figured we probably want to own it here regardless? I might be misunderstanding

rasendubi: The answer is correct for `Cow<str> -> String` conversion, however we're going from `Cow<str> -> FastStr` which may skip memory allocation for strings of up to 24 bytes (when converting from `Cow::Borrowed`).

Another important reason why we're using `.to_cow()` here is because it's the most performant way to get a string from Python with pyo3. Python 3.10 API introduced new functions to get the string reference directly (which would return `Cow::Borrowed` here) and for Python versions before that, it needs to copy the string ([source](https://github.com/PyO3/pyo3/blob/95949944554354cdbf3275daa03818a6f5b65c33/src/types/string.rs#L333-L348)).

eppo_core/src/eval/eval_assignment.rs
leoromanovsky: @rasendubi If the assignment base is already pre-computed what benefit is there to `cloning` it and making another copy, if the goal of this optimization is to reduce the memory allocated during evaluation? 

Is it because of Rust's ownership model? Can we just reference the value and create an `AssignmentEvent` by reference?

eppo_core/src/events.rs
```diff
@@ -1,79 +1,96 @@
...
-pub struct AssignmentEvent {
+pub struct AssignmentEventBase {
     /// The key of the feature flag being assigned.
-    pub feature_flag: String,
+    pub feature_flag: Str,
     /// The key of the allocation that the subject was assigned to.
-    pub allocation: String,
+    pub allocation: Str,
     /// The key of the experiment associated with the assignment.
     pub experiment: String,
```
leoromanovsky, line 25: Should `experiment` be an `Str` too?

eppo_core/src/eval/eval_assignment.rs
rasendubi: We're not making another copy because it's behind an `Arc` (atomic reference counter). When we're cloning the `Arc`, it just increases the reference counter without copying the data.

I believe it's similar to Swift, except it's just a normal/explicit type in Rust (it's not built-in)

eppo_core/src/events.rs
rasendubi: It doesn't need to be because it's never cloned. `String` is slightly better because `format!()` returns `String`

eppo_core/src/sharder.rs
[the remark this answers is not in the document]
schmit: TIL, that's awesome, love the implementation

eppo_core/src/events.rs
schmit: Why is experiment never cloned, but allocation is?

leoromanovsky (approved):
Really appreciate the detail you went into here to help us review. Sven, Aaron and I got together for 30 minutes and read through it.

The only downside to the optimizations could come up when we port this to clients, pre-computing the assignments could have a negative since for one user they will likely not trigger most flags. Something we will keep in mind for later.

eppo_core/src/events.rs
rasendubi: Because we clone `allocation` _from_ parsed UFC config but `experiment` is created for each split. After `AssignmentEventBase` is constructed, the whole event base is reused (via Arc)—we don't clone individual fields

rasendubi:
> The only downside to the optimizations could come up when we port this to clients, pre-computing the assignments could have a negative since for one user they will likely not trigger most flags. Something we will keep in mind for later.

This shouldn't be a huge issue as there's not a ton of work happening during the pre-compute phase (my guess is that compiling match rules is the most expensive step) and there's not much extra allocation happening as most data is shared and would be present in configuration anyway.

However, that's something I want to keep an eye on. e.g., benchmark precompute stage, so we could be more conscious about the pre-compute / evaluation tradeoff. If pre-compute becomes too huge, we may pre-compute it lazily per-flag (which also makes more sense for clients).

PR #42, opened 2024-10-11
fix(ruby): Enable ruby-sdk to be cross compiled ahead of time

felipecsl:
## Description
This will allow us to, at gem build time, precompile the ruby-sdk gem to all supported environments, so users don't need to have `cargo` installed to install the gem, skipping the whole "Building native extensions.  This could take a while..." annoyance

## Testing

[We support](https://github.com/Eppo-exp/ruby-sdk?tab=readme-ov-file#supported-ruby-versions) ruby versions 3.0 and above
```
cd ruby-sdk
bundle install
bundle exec rb-sys-dock -p <platform> --ruby-versions 3.3,3.2,3.1,3.0 --build
```

Also added Github action to cross compile as part of CI. We can add an upload step after, that part is still TBD

## References
* https://github.com/BoundaryML/baml/blob/canary/.github/workflows/build-ruby-release.reusable.yaml
* https://github.com/oxidize-rb/actions/blob/main/cross-gem/action.yml
* openssl issues https://github.com/openssl/openssl/discussions/24840
* https://github.com/rake-compiler/rake-compiler-dock/issues/122
* https://github.com/openssl/openssl/pull/25367#issuecomment-2338141250


felipecsl:
I got almost all of the main architectures working, except for `arm64-darwin` (that one has proven more challenging), however I doubt that anyone out there is running the ruby eppo sdk in prod on apple silicon :) I think we could get away with leaving that one aside for now and ship this as is. we can upload all individual binaries to rubygems and that should remove the need from users to have `cargo` installed at `gem install` time.

.github/workflows/ci.yml
```diff
@@ -1,4 +1,4 @@
-name: Test
+name: Ruby SDK
```
rasendubi, line 1: major: this is a general test workflow, not ruby-specific one. GitHub kinda messes up Actions tab on workflow renames so it's better to leave it as is

ruby-sdk/ext/eppo_client/extconf.rb
```diff
@@ -3,4 +3,4 @@
 require "mkmf"
 require "rb_sys/mkmf"
 
-create_rust_makefile("eppo_client")
+create_rust_makefile("eppo_client")
```
rasendubi, line 6: minor: accidentally removed trailing newline

ruby-sdk/ext/eppo_client/Cargo.toml
```diff
@@ -16,6 +16,7 @@ eppo_core = { version = "4.0.0" }
 log = { version = "0.4.21", features = ["kv_serde"] }
 magnus = { version = "0.6.4" }
 serde = { version = "1.0.203", features = ["derive"] }
+openssl = { version = "*", features = ["vendored"] }
```
rasendubi, line 19: I believe we don't need to vendor openssl on macOS and Windows because reqwest will use system's native library.

You can take a look at python SDK: https://github.com/Eppo-exp/rust-sdk/blob/427fed8fd54f3bdc7b69a9f2e198a72ef5ba4779/python-sdk/Cargo.toml#L18-L36

This should also fix macOS build.

.github/workflows/ruby-cross-compile.yml
rasendubi, line 1: minor: probably makes sense to rename this file to `ruby.yml` and move test/release here as well (so it's more aligned with `python.yml`)

.github/workflows/ruby-cross-compile.yml
```diff
@@ -0,0 +1,129 @@
...
+          : Compile gem
+          echo "Docker Working Directory: $(pwd)"
+          set -x
+
+          # Unfortunately we can't actually parallelize the Ruby versions
+          # because they get bundled into the same gem
+          rb-sys-dock \
+            --platform ${{ matrix._.platform }} \
+            --mount-toolchains \
+            --ruby-versions 3.3,3.2,3.1 \
```
rasendubi, line 127: 3.0 missing?

rasendubi (commented):
Awesome work!

How pre-built gems flow into ruby publish workflow?

.github/workflows/ruby-cross-compile.yml
```diff
@@ -0,0 +1,129 @@
+---
+name: Build Ruby gems
```
leoromanovsky, line 2: How does this fit in with ruby testing, such as:

```
 - name: Run tests
        run: npm run with-server test:ruby
```

Can this be ported and make this file the canonical ruby testing and building file?

leoromanovsky (commented):
🚀 🚀 🚀

.github/workflows/ci.yml
leoromanovsky: @rasendubi Do you want us to have a separate flow for each language? I see you created a `python.yml` alongside `ci.yml` and `publish.yml`.

rasendubi: I don't have a strong preference. Python workflow file was largely auto-generated, so it made sense to pull all python stuff into one file (including release). With a large portion of Ruby pushed to a new workflow, it makes sense to have a workflow per SDK (and remove `ci` and `publish` files later)

rasendubi:
oh, another caveat is that it tries to build the gems with `eppo_core` from crates.io. So it fails when `eppo_core` in worktree has bumped version

rasendubi:
So... Let's approach it in two steps:
1. I've opened #45 for changes in `eppo_core`. Let's merge them and release new `eppo_core`, so we can test this PR against released version.
2. After the core is released, we can continue working on workflow changes for Ruby (this PR). Unfortunately, we shall push all native gem compilation so it only happens during Ruby's release job (because in PR we want to test against local `eppo_core`, not the released version). I expect this to be fine because there's usually very little changes that would affect vendoring/cross-compilation and we have Python CI that builds very similarly.

felipecsl:
> Awesome work!
> 
> How pre-built gems flow into ruby publish workflow?

I'm gonna look into that next. There's a way to set up a workflow that publishes the gems for each platform after they are built. I'll look into that in a follow up PR if that's OK for you guys

.github/workflows/ruby.yml
```diff
@@ -0,0 +1,173 @@
...
+
+      - name: Build
+        run: bundle exec rake build
+        working-directory: ruby-sdk
+
+      - name: Run tests
+        run: npm run with-server test:ruby
+
+
+  cross_gems:
```
rasendubi, line 56: Before we can merge this, `cross_gems` needs to be configured to run on release only. Otherwise, it's going to fail a lot on PRs/main

rasendubi (commented):
Looks good and I'm glad it's passing for macOS as well now. Need to disable `cross_gems` for PRs/main before merging though

leoromanovsky (approved):
Approving but please extract `cross_gems` into own action that is triggered on `Release` before merging.

felipecsl:
cool, taking care of that now

PR #43, opened 2024-10-15
FF-3390 fix(python): fix typing information for Configuration.__init__

leoromanovsky:
🎫 Internal eppo ticket: `FF-3390`

## Observation

When using Python on version `4.0.1` (latest), unable to initialization bandit configuration from external payload.

👉  Only bandit keys that have a corresponding key in the model configuration are available in the `get_bandit_keys` function.

## Changes

* Augments the python configuration wrapper to match the parameters of the rust library.
* Adds unit test to verify the bandit configuration loading.
* Bumped `sdk-test-data` submodule to latest.


python-sdk/python/eppo_client/_eppo_client.pyi
```diff
@@ -6,7 +6,23 @@ def init(config: ClientConfig) -> EppoClient: ...
...
+    # In v5.0.0, we will add a named parameter to the constructor to make this
+    # more clear.
+    #
+    # Users will be able to initialize the configuration like this:
+    # 
+    # Configuration(flags_configuration=some_bytes, bandits_configuration=some_other_bytes)
+    #
+    # Changes:
+    # 
+    # def __init__(self, *, flags_configuration: bytes, bandits_configuration: bytes | None = None) -> None: ...
```
rasendubi, line 24: It actually already has this signature, so no need to wait for v5:
https://github.com/Eppo-exp/rust-sdk/blob/427fed8fd54f3bdc7b69a9f2e198a72ef5ba4779/python-sdk/src/configuration.rs#L22

python-sdk/tests/test_configuration.py
```diff
@@ -5,12 +5,94 @@
 
 from .util import init
 
+FLAGS_CONFIG = json.dumps({
+    "createdAt": "2024-09-09T10:18:15.988Z",
+    "environment": {"name": "test"},
+    "flags": {}
+}).encode('utf-8')
+
+FLAGS_CONFIG_WITH_BANDITS = ufc_with_bandits = json.dumps({
```
rasendubi, line 14: double-assignment here is confusing

rasendubi (approved):
Thanks for handling this!

rasendubi:
Just to clarify what's happening here, it's not that "configuration wrapper" didn't have the necessary parameter — it's just typing information that was out-of-sync (`.pyi` files are just type hints). There's no actual behavior change in this PR.

PR #44, opened 2024-10-15
chore: clarify release process

rasendubi:
We had a small incident with `python-sdk@4.0.2` tag pushed without updating the version strings so it tried to publish new artifacts into already released `python-sdk@4.0.1` on pypi. Luckily, none of artifacts were actually pushed because they already exist.

This incident went somewhat unnoticed because there were no GitHub release, just a tag. So I only noticed this when trying to push a new release and found that the tag already exists.

I have rectified this by deleting the old `python-sdk@4.0.2` tag that did not have effect and created a proper release.

This PR is a follow-up and:
- Modifies Python workflow to require a GitHub release, so releases are more visible.
- Adds documentation on the release process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,3 +4,22 @@
 
 1. Install [rustup](https://rustup.rs/).
 2. Install Ruby using your preferred package manager.
+
+### Release process
+
+To release a new version of SDK:
+1. Make sure that version strings have been updated:
+   - Eppo core: [eppo_core/Cargo.toml](eppo_core/Cargo.toml)
+   - Rust: [rust-sdk/Cargo.toml](rust-sdk/Cargo.toml)
+   - Python: [python-sdk/Cargo.toml](python-sdk/Cargo.toml)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -4,3 +4,22 @@
 
 1. Install [rustup](https://rustup.rs/).
 2. Install Ruby using your preferred package manager.
+
+### Release process
+
+To release a new version of SDK:
+1. Make sure that version strings have been updated:
+   - Eppo core: [eppo_core/Cargo.toml](eppo_core/Cargo.toml)
```
leoromanovsky, line 12: My understanding is this should be treated with its own semver tree; for example breaking changes in eppo core might not be breaking changes up stream?

sameerank (approved):
Thank you! I would also appreciate more notes on local env setup (in a future PR is fine), e.g. 
* what do I need to do before I run `npm run with-server test:python` to see passing tests?
* can I create a distribution package for python to install locally and try out before publishing to pypi?

README.md
rasendubi: Yes, all components have independent versioning

PR #46, opened 2024-10-16
fix: flatten "base" in AssignmentEvent

rasendubi:


eppo_core/src/events.rs
```diff
@@ -38,6 +38,7 @@ pub struct AssignmentEventBase {
 #[derive(Debug, Clone, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct AssignmentEvent {
+    #[serde(flatten)]
```
rasendubi, line 41: "flatten" unpacks a nested structure into parent when serializing

python-sdk/tests/test_assignment_logger.py
```diff
@@ -27,3 +29,23 @@ def test_has_log_assignment():
...
+            nonlocal event
+            event = assignment_event
+
+    client = init("ufc", assignment_logger=MyAssignmentLogger())
+    result = client.get_string_assignment(
+        "regex-flag", "alice", {"email": "alice@example.com"}, "default"
+    )
+    assert result == "partial-example"
+
+    assert event["allocation"] == "partial-example"
```
leoromanovsky, line 48: Does this demonstrate that nested elements, like `base` are no longer present?

rasendubi: yes

leoromanovsky (approved):

felipecsl (approved):

PR #50, opened 2024-10-17
FF-3392 feat(python): add Configuration.get_bandits_configuration

rasendubi:
Expose `get_bandits_configuration` in Python SDK. Ruby SDK already has it exposed. Rust SDK doesn't currently expose Configuration API.

Example of using new API to export configuration to files:
```py
import os

import eppo_client

client = eppo_client.init(
    eppo_client.ClientConfig(
        api_key=os.environ["EPPO_API_KEY"],
        assignment_logger=eppo_client.AssignmentLogger(),
    )
)

client.wait_for_initialization()

if configuration := client.get_configuration():
    flags = configuration.get_flags_configuration()
    with open("flags.json", "wb") as f:
        f.write(flags)

    if bandits := configuration.get_bandits_configuration():
        with open("bandits.json", "wb") as f:
            f.write(bandits)
```

@aarsilv it's not a static method as we discussed but I find it quite straightforward to use anyway. I would like to keep API surface smaller for now (it's easier to add new API later than change/remove existing one).

### DIFF
--- a/python-sdk/Cargo.toml
+++ b/python-sdk/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "eppo_py"
-version = "4.0.3"
+version = "4.1.0"
 edition = "2021"
 publish = false
 


.github/workflows/python.yml
```diff
@@ -82,7 +82,7 @@ jobs:
           set -ex
           python3 -m venv .venv
           source .venv/bin/activate
-          pip install eppo-server-sdk --find-links dist --force-reinstall
+          pip install eppo-server-sdk --no-index --find-links dist --force-reinstall
```
rasendubi, line 85: Without `--no-index`, the CI was [randomly](https://github.com/Eppo-exp/rust-sdk/actions/runs/11388672269) selecting either local package or one published on pypi. See https://github.com/Eppo-exp/rust-sdk/pull/50/commits/36f083b2b5d78e69a0167c6fa8916ba03892573c

leoromanovsky (approved):
Perfect!

python-sdk/src/configuration.rs
```diff
@@ -68,6 +72,23 @@ impl Configuration {
...
+            .as_ref()
+            .map(|it| serde_json::to_vec(it).map(Cow::Owned))
+            .transpose()
+            .map_err(|err| {
+                // This should normally never happen.
+                PyRuntimeError::new_err(format!(
+                    "failed to serialize bandits configuration: {err:?}"
+                ))
+            })
+    }
```
leoromanovsky, line 91: I observed that the implementation for `get_flags_configuration` and `get_bandits_configuration` is different as it related to the rust borrow checker and wanted to learn more about why.

Screenshot 2024-10-17 at 3 40 00 PM

Do you have more to add about your decision?

.github/workflows/python.yml
aarsilv: kind of wild default behavior 😅

python-sdk/tests/test_configuration.py
```diff
@@ -116,6 +115,19 @@ def test_init_invalid_format(self):
...
+            flags_configuration=FLAGS_CONFIG_WITH_BANDITS,
+            bandits_configuration=BANDITS_MODEL_CONFIG,
+        )
+
+        bandits_configuration = configuration.get_bandits_configuration()
+
+        # JSON parsing is an internal detail and is not a public
+        # guarantee. We're using it here because serialization order
+        # of bandits is not guaranteed.
+        assert json.loads(bandits_configuration) == json.loads(BANDITS_MODEL_CONFIG)
```
aarsilv, line 129: assertion seems sound

python-sdk/src/configuration.rs
```diff
@@ -68,6 +72,23 @@ impl Configuration {
     fn get_flags_configuration(&self) -> Cow<[u8]> {
         Cow::Borrowed(self.configuration.flags.to_json())
     }
+
+    /// Return bytes representing bandits configuration.
+    ///
+    /// It should be treated as opaque and passed on to another Eppo client for initialization.
+    fn get_bandits_configuration(&self) -> PyResult<Option<Cow<[u8]>>> {
+        self.configuration
+            .bandits
```
aarsilv, line 81: Possibly stupid question: is this the handle the downloaded (raw) bytes or are we re-serializing our internal representation?

(I'd prefer the former--just saving off and then returning when requested what we initialized with)

python-sdk/src/configuration.rs
rasendubi: The main difference is that flags configuration holds the server response as is, so we return a reference to it (`&[u8]`) without transferring ownership.

For bandits, configuration is serialized every time into a new `Vec<u8>`, and we transfer ownership to the caller (so they free the memory when it's no longer needed).

The reason we wrap results in `Cow` (instead of having functions return `&[u8]` and `Vec<u8>`) is to have return value [converted](https://pyo3.rs/v0.22.5/conversions/tables#returning-rust-values-to-python) to Python's `bytes` (`&[u8]` and `Vec<u8>` are converted to a list of numbers).

python-sdk/src/configuration.rs
rasendubi: It is re-serializing internal representation.

The flags configuration has been updated to hold the server response as is (it was actually done before our discussion took place) but bandits configuration has not been updated yet.

leoromanovsky: Thanks for explaining that, makes sense.

PR #53, opened 2024-10-22
fix(ruby): Require shared lib from correct subdir

felipecsl:
rake-compiler puts the shared lib in a subdirectory of `eppo_client` when we build with multiple ruby target versions, so we need to load the lib from the correct subdir.

Again, shamelessly stolen from wasmtime.

Also fixed the CI step so it runs properly (it wasnt)

### DIFF
--- a/ruby-sdk/README.md
+++ b/ruby-sdk/README.md
@@ -23,5 +23,6 @@ Make sure you remove the override before updating `Cargo.lock`. Otherwise, the l
 
 ## Releasing
 
-* Bump versions in `ruby-sdk/lib/eppo_client/version.rb` and `ruby-sdk/ext/eppo_client/Cargo.toml`.
-* Run `cargo update --workspace --verbose` to update `Cargo.lock` and `Gemfile.lock`.
+* Bump versions in `ruby-sdk/lib/eppo_client/version.rb` and `ruby-sdk/ext/eppo_client/Cargo.toml`
+* Run `cargo update --workspace --verbose` to update `Cargo.lock`
+* Run `bundle` to update `Gemfile.lock`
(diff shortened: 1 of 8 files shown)


ruby-sdk/README.md
```diff
@@ -23,5 +23,6 @@ Make sure you remove the override before updating `Cargo.lock`. Otherwise, the l
 
 ## Releasing
 
-* Bump versions in `ruby-sdk/lib/eppo_client/version.rb` and `ruby-sdk/ext/eppo_client/Cargo.toml`.
-* Run `cargo update --workspace --verbose` to update `Cargo.lock` and `Gemfile.lock`.
+* Bump versions in `ruby-sdk/lib/eppo_client/version.rb` and `ruby-sdk/ext/eppo_client/Cargo.toml`
+* Run `cargo update --workspace --verbose` to update `Cargo.lock`
+* Run `bundle` to update `Gemfile.lock`
```
leoromanovsky, line 28: The developer needs to do this and check on it?

leoromanovsky (approved):
Any steps to demonstrate a more favorable build?

ruby-sdk/README.md
felipecsl: correct. right now these are manual steps but we can certainly automate this in the future

felipecsl:
> Any steps to demonstrate a more favorable build?

The new `Smoke gem install` should fail now upon cutting a new release if there are any issues 🤞🏽

felipecsl:
Looks like it did indeed fail https://github.com/Eppo-exp/rust-sdk/actions/runs/11456821053/job/31875784240 but looks like a false negative, just forgot to add quotes around the `grep` argument

PR #58, opened 2024-10-22
disable builds on macos

leoromanovsky:
Failing with cryptic error - in the interest of delivering a build to a customer on a linux environment I want to unblock this.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -211,45 +211,45 @@ jobs:
           npm ci
           npm run with-server test:python
 
-  # macos:
-  #   runs-on: ${{ matrix.platform.runner }}
-  #   strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/python.yml
```diff
@@ -211,45 +211,45 @@ jobs:
...
-          npm ci
-          npm run with-server test:python
+  # macos:
+  #   runs-on: ${{ matrix.platform.runner }}
+  #   strategy:
+  #     matrix:
+  #       platform:
+  #         - runner: macos-12
+  #           target: x86_64
+  #         - runner: macos-14
```
felipecsl, line 221: the only target that seems to be failing is macos-14 so we can just comment this part out

felipecsl (approved):

PR #67, opened 2024-11-09
chore: generate rust-sdk github action with build.rs

leoromanovsky:
**Observation**

* Developed github actions is tedious and slow - validating correctness and behavior has to happen by pushing to remote and waiting for the test to run.
* The yaml syntax needs to be learned.

**Desired behavior**

* Local validation
* Compiler-guided rule building

**Changes proposed**

Use the `gh_workflow` crate in `build.rs` to automatically keep the github workflow files updated.

Rust's buildscripts (https://doc.rust-lang.org/cargo/reference/build-scripts.html) run on compilation; this PR generates no new work for developers.

**Testing**

* the `ci.yml` file should be functionally equivalent
* CI runs - https://github.com/Eppo-exp/eppo-multiplatform/actions/runs/11791022187/job/32842408815?pr=67


rust-sdk/Cargo.toml
```diff
@@ -21,3 +21,8 @@ name = "simple"
 [dev-dependencies]
 chrono = "0.4.38"
 env_logger = { version = "0.11.3", features = ["unstable-kv"] }
+
+[build-dependencies]
+gh-workflow = "0.3.0"
```
leoromanovsky, line 26: A dependency in this package has a MSRV of `1.75` which blocks us from using it: https://github.com/JelteF/derive_more/pull/389

I will fiddle around with it to see if we can overcome the blocker.

rasendubi:
I'm actually not a fan of this approach.

It might be cool but it doesn't seem to solve any real issue:
- It doesn't help you to get away from learning yaml and github actions. It just adds a layer of indirection. 
  - This is especially true as we have multiple workflow and only one is auto-generated — you still need to know and write yamls. *And* we have people changing wokrflows who don't have rust knowledge: #73 is changing exactly this file.
- It doesn't help with debugging workflow besides syntax. There doesn't seem to be a command to run it locally — you still need to push your changes to a branch and see if CI passes

and adds a few new:
- Idiomatically, `build.rs` is designed to configure the compiler and generate source files (which are not commited). It is not designed for generating arbitrary files. One outcome of that is that `build.rs` runs for *users* of our library when they install it.
- As `yml` file is now generated, this adds a bunch of downsides like possible sync issues, people accidentally modifying generated files, slowing down editing (as you have to run it through an additional tool), etc.
- Github actions yaml is more widely known and this is a new tool that everybody has to learn

PR #69, opened 2024-11-16
feat: fastly compute service to perform assignments (FF-3552)

leoromanovsky:
## Motivation

Eppo is launching a server-based assignment service hosted on the Fastly Compute platform. This is an edge-product that makes 

Design details available in the internal doc.

## PR goals

* Create a service that is able to respond to `POST /assignments` with a `subjectKey` and `subjectAttributes` and return assignments for active allocations.

## PR non-goals

This work does not include integration testing or CICD integration beyond compilation. This functionality will be added before a `1.0.0` release in the coming weeks.

## Description

* Service responds to requests on `GET /health` and `POST /assignments?apiKey` - the `apiKey` format is re-used for legacy reasons although the marketing copy is `SDK Key`.
* The incoming `apiKey` is converted to a hash and used to look up items in the attached KV store. An external process populates the KV store on Flag changes.
* The handler performs assignment across all available flags in series and returns the ones with success evaluation. Any errors in evaluation are logged and the flags are not included in the response.


fastly-edge-assignments/src/handlers/assignments.rs
```diff
@@ -0,0 +1,164 @@
...
+            .into_iter()
+            .filter_map(|(key, opt_assignment)| {
+                opt_assignment.map(|assignment| (key, assignment.value))
+            })
+            .collect::<HashMap<_, _>>(),
+        Err(e) => {
+            // If we encounter an error in any of the assignments, return an internal server error.
+            //
+            // If any of the assignments produces an error during evaluation, the collection will short-circuit and return that first error.
+            // It won't continue processing the remaining assignments.
```
rasendubi, line 141: nit: short-circuiting is not exactly true because all assignments have been evaluated by the time `get_subject_assignments()` returned (because it materialized results to an interim `HashMap` that is converted to a new iterator with `.into_iter()`).

It's slightly better to collect into `Result` directly (without in-between HashMap), so it would short-circuit properly.

leoromanovsky: Good observation thank you - apologize for that; was working with chatgpt to figure this out. I'll make that change, cool!

Makefile
```diff
@@ -1,28 +1,47 @@
...
+	cargo build --workspace --exclude $(FASTLY_PACKAGE)
+
+# Run tests for the entire workspace excluding the `fastly-edge-assignments` package
+.PHONY: workspace-test
+workspace-test:
+	cargo test --workspace --exclude $(FASTLY_PACKAGE)
+
+# Build only the `fastly-edge-assignments` package for WASM
+.PHONY: fastly-edge-assignments-build
+fastly-edge-assignments-build:
```
leoromanovsky, line 40: Are these changes helpful? I got tired of putting the various parameters in the right place.

fastly-edge-assignments/fastly.toml
```diff
@@ -0,0 +1,17 @@
...
+manifest_version = 3
+
+[scripts]
+build = "cargo build --bin fastly-edge-assignments --release --target wasm32-wasi --color always"
+
+[local_server]
+  [local_server.kv_stores]
+    [[local_server.kv_stores.edge-assignment-kv-store]]
+      key = "ufc-by-sdk-key-token-hash-V--77TScV5Etm78nIMTSOdiroOh1__NsupwUwsetEVM"
+      file = "../sdk-test-data/ufc/flags-v1.json"
```
leoromanovsky, line 17: this instructs the `fastly serve` process to populate the local in-memory KV store with one key, a UFC configuration at a precomputed hash. See the unit test in this repo for the associate SDK key to use.

fastly-edge-assignments/src/handlers/assignments.rs
```diff
@@ -0,0 +1,241 @@
...
+    variation_type: String,
+    variation_value: serde_json::Value,
+    extra_logging: HashMap<String, serde_json::Value>,
+    do_log: bool,
+}
+
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+struct AssignmentsResponse {
+    created_at: i64,
```
rasendubi, line 50: major: `i64` deviates from the tech design and how we usually represent timestamps.

We could use `chrono::DateTime<chrono::Utc>` and it would serialize to ISO timestamp:
```suggestion
    created_at: chrono::DateTime<chrono::Utc>,
```

fastly-edge-assignments/src/handlers/assignments.rs
```diff
@@ -0,0 +1,241 @@
...
+    format: AssignmentFormat,
+    environment: Environment,
+    flags: HashMap<String, FlagAssignment>,
+}
+
+const KV_STORE_NAME: &str = "edge-assignment-kv-store";
+const SDK_KEY_QUERY_PARAM: &str = "apiKey"; // For legacy reasons this is named `apiKey`
```
rasendubi, line 57: minor: given this is a completely new service, I think we can go with `sdkKey` as there are no legacy clients?

fastly-edge-assignments/src/handlers/assignments.rs
```diff
@@ -0,0 +1,241 @@
...
+    format!("ufc-by-sdk-key-token-hash-{}", token_hash)
+}
+
+fn token_hash(sdk_key: String) -> String {
+    let mut hasher = Sha256::new();
+    hasher.update(sdk_key.as_bytes());
+    base64_url::encode(&hasher.finalize())
+}
+
+pub fn handle_assignments(mut req: Request) -> Result<Response, Error> {
```
rasendubi, line 72: minor: this function has potential to be split into multiple helper functions, which is currently hindered by the fact that error responses are constructed inline. If we have an explicit error type, that would make splitting much easier and the main body more pleasant and straightforward

fastly-edge-assignments/src/main.rs
```diff
@@ -0,0 +1,13 @@
...
+use fastly::{Error, Request, Response};
+
+#[fastly::main]
+fn main(req: Request) -> Result<Response, Error> {
+    match (req.get_method(), req.get_path()) {
+        (&Method::POST, "/assignments") => handlers::handle_assignments(req),
+        (&Method::GET, "/health") => handlers::handle_health(req),
+        _ => Ok(Response::from_status(StatusCode::NOT_FOUND).with_body_text_plain("Not Found")),
+    }
+}
```
sameerank, line 13: I was hitting a CORS error running the SDK in the browser

Screenshot 2024-11-22 at 12 01 52 AM

I think this is potentially a viable place to add CORS headers. If you agree, then consider adding them here for the preflight requests and responses. There are other places they could be configured though, e.g. terraform, so this is just a suggestion

fastly-edge-assignments/src/handlers/assignments.rs
leoromanovsky: Yea I was thinking about this too - I certainly want to split out functions where reasonable but only if we think there are multiple consumers. Do you have some suggestions?

fastly-edge-assignments/src/handlers/assignments.rs
leoromanovsky: I want to do this as well - @sameerank used our existing javascript client to access this service to setup a connection for our eppo.cloud QA environment so we can start collect telemetry on real-world use. https://github.com/Eppo-exp/js-sdk-common/pull/137

I have a couple of goals to evolve the network architecture of this service:

* Serve it behind our existing CDN (with no caching) to be able to share the same domain name and SSL certificate
* Since Sameeran developed the new client separately from the old we have a lot of control over the query parameters and should make this change; or even better, pass the auth information in a `Authorization` header.

fastly-edge-assignments/src/handlers/assignments.rs
```diff
@@ -0,0 +1,241 @@
...
+    name: String,
+}
+
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+struct FlagAssignment {
+    allocation_key: String,
+    variation_key: String,
+    variation_type: String,
+    variation_value: serde_json::Value,
```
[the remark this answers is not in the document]
leoromanovsky: `ValueWire` is `pub(crate)` - do you have a suggestion for how the edge function can access it?

eppo_core/src/precomputed_assignments.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// Additional user-defined logging fields for capturing extra information related to the
+    /// assignment.
+    #[serde(flatten)]
+    pub extra_logging: HashMap<String, String>,
+    pub do_log: bool,
+}
+
+impl FlagAssignment {
+    pub fn try_from_assignment(assignment: Assignment) -> Option<Self> {
+        // WARNING! There is a problem here. The event is only populated for splits
```
leoromanovsky, line 35: @rasendubi @sameerank I believe this is a problem as described here: https://github.com/Eppo-exp/eppo-multiplatform/blob/main/eppo_core/src/ufc/compiled_flag_config.rs#L231

During compilation, an optimization step to reduce the configuration, `event` is only populated for splits with `do_log == True`. 

A possible solution is to produce a configuration that is "not optimized" for use in this function so that we can access `do_log`

eppo_core/src/ufc/models.rs
```diff
@@ -18,6 +18,7 @@ pub type Timestamp = chrono::DateTime<chrono::Utc>;
 pub(crate) struct UniversalFlagConfigWire {
     /// When configuration was last updated.
     pub created_at: Timestamp,
+    pub format: AssignmentFormat,
```
leoromanovsky, line 21: heads up @sameerank I added a new key to UFC parsing. I see it is being returned now always so I feel it is safe to make it required.

fastly-edge-assignments/src/handlers/assignments.rs
leoromanovsky: Ok I see what you mean - ideally we return explicit errors and this handle just is responsible to rendering those into various http codes and messages?

leoromanovsky:
@sameerank I'm going to merge this PR even though we still have some feedback from Oleksii and improvements I pointed out in the comments; the `do_log` issue is most pressing to figure out.

eppo_core/src/precomputed_assignments.rs
```diff
@@ -0,0 +1,86 @@
...
+            allocation_key: event.base.allocation.clone(),
+            variation_key: event.base.variation.clone(),
+            variation_type: assignment.value.variation_type(),
+            variation_value: assignment.value.variation_value(),
+            extra_logging: event
+                .base
+                .extra_logging
+                .iter()
+                .map(|(k, v)| (k.clone(), v.clone()))
+                .collect(),
```
rasendubi, line 55: same as:
```rust
extra_logging: event.base.extra_logging.clone(),
```

fastly-edge-assignments/src/handlers/assignments.rs
```diff
@@ -0,0 +1,175 @@
...
+                }
+            }
+        })
+        .collect::<HashMap<_, _>>();
+
+    // Create the response
+    let assignments_response = PrecomputedAssignmentsServiceResponse::from_configuration(
+        configuration,
+        subject_assignments,
+    );
```
rasendubi, line 146: major: computing subject assignments should be part of core. super-weird to have request/response formats in the core but the way to calculate assignment not in the core

eppo_core/src/precomputed_assignments.rs
```diff
@@ -0,0 +1,86 @@
+use crate::ufc::{Assignment, AssignmentFormat, Environment, VariationType};
+use crate::{Attributes, Configuration, Str};
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::sync::Arc;
+
+// Request
+#[derive(Debug, Deserialize)]
+pub struct PrecomputedAssignmentsServiceRequestBody {
```
rasendubi, line 9: major: `eppo_core` shall be unconcerned with network request format

eppo_core/src/ufc/models.rs
rasendubi, line 21: minor: ideally, this field should be optional because there could exist configurations without it (remember that users may be storing and passing configurations out of band now)

eppo_core/src/precomputed_assignments.rs
leoromanovsky: I still think its a problem because we want the precomputed server to return evaluations for all flags and `do_log` is true or false depending on if the user wants them log, but they still need to be present. The flag will be present in the response which changes the behavior of the SDK / application instead of now the flag will not be present and the default value will be used. Am I understanding that correctly?

eppo_core/src/ufc/models.rs
leoromanovsky: https://github.com/Eppo-exp/eppo-multiplatform/pull/76

PR #70, opened 2024-11-17
feat: add `get_subject_assignments` method to eppo_core to compute al…

leoromanovsky:
…l assignments for a given subject against the cached flags (FF-3571)

## Motivation

To support a bulk assignment function, we need to be able to evaluate all flags for a given `(subject, attributes)` tuple. 

My goal is to evaluation against all flags and produce an assignment or evaluation error without early exiting. This allows the callers to chose their desired behavior: they might decide to throw an exception if any of the values are an error or just return evaluations that succeeded.

## Description

The function API is `HashMap<String, Result<Option<Assignment>, EvaluationError>>` with the key being the `flagKey`. This provides an evaluation for each flag and a positive or exceptional result.

Using available keys in the `compiled` UFC configuration to loop over existing evaluation. I considered an alternative approach where a vector of `flagKey` would be provided but this seemed like added complication since the function has access to the configuration.

I considered using `https://docs.rs/rayon/latest/rayon/` to achieve parallelism but my gut feeling is the overhead will not be worth it as usually there are only a few hundred flags and each unit of work is very fast, but I am open to further refinement through benchmarking.

## Help from reviewers

* Does the API feel like it achieves the desired goal?
* Is looping over the compiled flags feel hacky?

### DIFF
--- a/eppo_core/src/eval/eval_assignment.rs
+++ b/eppo_core/src/eval/eval_assignment.rs
@@ -129,13 +129,13 @@ pub(super) fn get_assignments_with_visitor<V: EvalAssignmentVisitor>(
                 let result = get_assignment_with_visitor(
                     Some(&config),
                     visitor,
-                    flag_key.as_str(),
+                    flag_key,
                     subject_key,
                     subject_attributes,
                     None,
                     now,
                 );
-                (flag_key.to_string(), result)
+                (flag_key.clone(), result)
             })
             .collect()
     })


eppo_core/src/eval/eval_assignment.rs
```diff
@@ -91,6 +91,56 @@ pub fn get_assignment_details(
...
+    now: DateTime<Utc>,
+) -> HashMap<String, Result<Option<Assignment>, EvaluationError>> {
+    configuration.map_or_else(HashMap::new, |config| {
+        config
+            .flags
+            .compiled
+            .flags
+            .keys()
+            .map(|flag_key| {
+                let result = get_assignment_with_visitor(
```
aarsilv, line 129: Pardon my lack of rust knowledge / memory of this codebase, but assuming graceful mode is on is there a top-level try catch here so an error getting one assignment won't tank all? (Or do we want it to tank all? Thinking not)

leoromanovsky: All questions welcome thanks for bringing it up! This is not the public method available through the rust sdk so the graceful mode is not relevant - `eppo_core` is an internal package.

The `Result` enum in rust is the canonical way to return a tuple of `(desired, exceptional)` behavior which I am employing here:

The return is `HashMap<String, Result<Option<Assignment>, EvaluationError>>` and evaluation will happen against all flags - the return will be each flag will have a `Result`

My thinking then is the caller can decide whether what to do if any fail. In the edge function I am writing it will return an HTTP exception if any assignment fails - https://github.com/Eppo-exp/eppo-multiplatform/pull/69/files#diff-ff0287fa81e178dd11d342cf54ba1519848b1a29bf0b06192fc0aaec64d0120bR138

rasendubi:
Is there anything that prevents you from looping outside of eppo core? i.e., is there any benefit to placing this code inside the core?

e.g., one benefit to placing it outside is that you could early-return in case of error (by collecting to `Result<HashMap<key, value>>` instead of `HashMap<key, Result<value>>`).

eppo_core/src/eval/eval_assignment.rs
rasendubi: A couple of notes and nits.

Rust doesn't have exception, so no try-catch, and no "non-graceful" mode. You can usually tell where the error boundary lies from the return type (where `Result` is). `HashMap<String, Result<Option<Assignment>, EvaluationError>>` for this functions means each hash map value would be its own result.

For this use case, it's actually better to merge this code with the calling site, so it could control that boundary. Right now, it continues to evaluate all assignments even if the first one failed. For the edge function, it's better to short-circuit and stop evaluating the rest of the flags if we're going to return error anyway.

nit: `Result` is not a tuple because tuples contain **both** values simultaneously. Enums contain only one of the two types (either-or). Mathematically, `struct`s behave as `AND` or `*` (multiplies the number of available values in the type), and enums in Rust are `OR` or `+` (number of available values is the sum of two domains). Which is why you may see this feature called "algebraic data types."

rasendubi: Thinking about it now, I would like to challenge the short-circuiting behavior. We usually default to returning default value if evaluation failed for any reason. It seems strange that we decide to abort evaluation for edge function and return a global error for all flags, which could cause SDKs to error out on all evaluations even if the failed flag is never used.

leoromanovsky:
> Is there anything that prevents you from looping outside of eppo core? i.e., is there any benefit to placing this code inside the core?
> 
> e.g., one benefit to placing it outside is that you could early-return in case of error (by collecting to `Result<HashMap<key, value>>` instead of `HashMap<key, Result<value>>`).
> 
> I would like to push the whole creation of precomputed configuration inside the core, so SDKs could issue precomputed configs in the future. This would also helps SDKs to parse precomputed configuration as the data model is already in core, etc.

Not a real "blocker" I just thought it would be useful in `eppo_core`. For now I can keep it in the assignment function, which is more convenient anyways to iterate on it quickly:

```
    let configuration = Configuration::from_server_response(ufc_config, None);
    let flag_keys = configuration.flag_keys();
    let configuration_store = ConfigurationStore::new();
    configuration_store.set_configuration(Arc::new(configuration));
    let evaluator = Evaluator::new(EvaluatorConfig {
        configuration_store: Arc::new(configuration_store),
        sdk_metadata: SdkMetadata {
            name: "fastly-edge-assignments",
            version: "0.1.0",
        },
    });

    flag_keys.iter().for_each(|key| {
        let assignment = evaluator.get_assignment(
            key,
            subject_key,
            &subject_attributes,
            None,
        );
        // ..
    });
```

eppo_core/src/eval/eval_assignment.rs
leoromanovsky: Yea your thinking makes sense @rasendubi - we could deliver as many flags as we have available which does seem better than none.

To support this should I keep the return type as `HashMap<String, Result<Option<Assignment>, EvaluationError>>` to be able to collect success and failures independently? I assume if I wrap the entire thing in a `Result` that is more of the binary "everything failed or everything was successful".

leoromanovsky:
Going to close this PR and continue development for this functionality in the new service. Continued discussions are welcome.

PR #72, opened 2024-11-21
FF-3593 feat(ruby): allow configuring log level from config

sameerank:
We need a way to adjust the log level in rust from ruby.

Currently, it appears that the log level is set to debug, which is too verbose.

https://github.com/Eppo-exp/eppo-multiplatform/blob/9d491dc04e32fba2c79987404eb68c8c99171fca/ruby-sdk/ext/eppo_client/src/lib.rs#L17

I propose moving it up to the `info` level and allowing it to be adjusted with an additional argument to the ruby config. This is just a proposal; another approach would be to automatically match the log level configured in the rails app, since that is in many cases where this sdk would be running.


ruby-sdk/lib/eppo_client/config.rb
```diff
@@ -6,14 +6,15 @@
...
+    attr_reader :api_key, :assignment_logger, :base_url, :poll_interval_seconds, :poll_jitter_seconds, :log_level
 
-    def initialize(api_key, assignment_logger: AssignmentLogger.new, base_url: EppoClient::Core::DEFAULT_BASE_URL, poll_interval_seconds: EppoClient::Core::DEFAULT_POLL_INTERVAL_SECONDS, poll_jitter_seconds: EppoClient::Core::DEFAULT_POLL_JITTER_SECONDS, initial_configuration: nil)
+    def initialize(api_key, assignment_logger: AssignmentLogger.new, base_url: EppoClient::Core::DEFAULT_BASE_URL, poll_interval_seconds: EppoClient::Core::DEFAULT_POLL_INTERVAL_SECONDS, poll_jitter_seconds: EppoClient::Core::DEFAULT_POLL_JITTER_SECONDS, initial_configuration: nil, log_level: "info")
       @api_key = api_key
       @assignment_logger = assignment_logger
       @base_url = base_url
       @poll_interval_seconds = poll_interval_seconds
       @poll_jitter_seconds = poll_jitter_seconds
+      @log_level = log_level
```
sameerank, line 17: I think it would be helpful to have a few unit tests for the log levels

rasendubi: It's a bit hard to test in general because apps usually don't have access to their stderr

sameerank (commented):
I initialized the client in irb and confirmed that the the log levels work as expected. Thanks! Looks like I can't approve my own PR though

Screenshot 2024-11-21 at 5 06 44 PM

Please self-approve if you can

rasendubi (approved):

PR #75, opened 2024-11-25
Create fastly-edge-assignments integration tests

leoromanovsky:
## Observation

Moving beyond the M1 for this project to a trusted tool, we need to add integration and unit tests.

A few constraints on testability is that we are compiling a `wasm` target and deploying to the fastly platform.

The purpose of this PR is to create the scaffolding for future tests and not an exhaustive suite.

## Description

For integration testing I am choosing to use the `fastly` cli which comes with a rust build program; this is the same process we use for building the deployable package.

[Viceroy](https://github.com/fastly/Viceroy) will let us locally create fastly services; we already use it for running the local environment with a pre-populated KV store entry.

`cargo-nextest` is a test runner recommended by fastly.

I followed the fastly guides (https://www.fastly.com/blog/how-to-write-rust-unit-tests-for-your-compute-edge-application).


fastly-edge-assignments/src/main.rs
```diff
@@ -3,8 +3,14 @@ mod handlers;
 use fastly::http::{Method, StatusCode};
 use fastly::{Error, Request, Response};
 
-#[fastly::main]
-fn main(req: Request) -> Result<Response, Error> {
+fn main() -> Result<(), Error> {
+    let ds_req = Request::from_client();
```
leoromanovsky, line 10: This is required scaffolding to run integration tests.

.github/workflows/fastly-edge-assignments.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Setup Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+          target: wasm32-wasi
+
+      # Install tools only if not cached
+      - name: Install Tools
```
leoromanovsky, line 49: this step takes ~7 min so I added it to the github cache

.github/workflows/fastly-edge-assignments.yml
```diff
@@ -0,0 +1,66 @@
+name: Fastly Edge Assignments
+
+on:
+  push:
+    branches:
+      - main
+  release:
+    types: [published]
```
rasendubi, line 8: Releases have a commit on `main`, so having a `release` trigger here will just execute the same workflow twice.

It's also going to be executed on *all* releases (core, rust, ruby, python sdks)

Suggest dropping this trigger:
```suggestion
```

rasendubi (approved):

PR #76, opened 2024-11-25
Make UniversalFlagConfigWire.format optional

leoromanovsky:
Feedback from: https://github.com/Eppo-exp/eppo-multiplatform/pull/69#discussion_r1856974637

### DIFF
--- a/eppo_core/src/ufc/models.rs
+++ b/eppo_core/src/ufc/models.rs
@@ -18,7 +18,8 @@ pub type Timestamp = chrono::DateTime<chrono::Utc>;
 pub(crate) struct UniversalFlagConfigWire {
     /// When configuration was last updated.
     pub created_at: Timestamp,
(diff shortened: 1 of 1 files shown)


eppo_core/src/ufc/models.rs
```diff
@@ -18,7 +18,8 @@ pub type Timestamp = chrono::DateTime<chrono::Utc>;
 pub(crate) struct UniversalFlagConfigWire {
     /// When configuration was last updated.
     pub created_at: Timestamp,
-    pub format: AssignmentFormat,
+    #[serde(default)]
```
rasendubi, line 21: so we don't serialize `null`s:
```suggestion
    #[serde(default, skip_serializing_if = "Option::is_none")]
```

rasendubi (approved):

PR #77, opened 2024-11-25
chore: upgrade deprecated macos-12 builds to macos-latest

leoromanovsky:
## Observation

`macos-12` runner is deprecated and has stopped working:

Screenshot 2024-11-25 at 3 19 39 PM

## Changes

Bumped to `macos-13` for `x86_64` and restored `aarch64` build on `macos-14`

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -216,10 +216,10 @@ jobs:
     strategy:
       matrix:
         platform:
-          - runner: macos-12
+          - runner: macos-13
             target: x86_64
-          # - runner: macos-14
-          #   target: aarch64
+          - runner: macos-14
+            target: aarch64
     steps:
       - uses: actions/checkout@v4
         with:


.github/workflows/python.yml
```diff
@@ -23,45 +23,36 @@ permissions:
...
-            target: aarch64
-          - runner: ubuntu-latest
-            target: armv7
-          - runner: ubuntu-latest
-            target: s390x
-          - runner: ubuntu-latest
-            target: ppc64le
+        runner: [ubuntu-latest]
+        target: [x86_64, x86, aarch64, armv7, s390x, ppc64le]
+        python-version: ["3.12"]
```
leoromanovsky, line 32: DRYing - only testing against latest python, increasing the matrix size seems like over-kill

also I couldn't figure out how to refactor this block to be the same for the entire file

.github/workflows/python.yml
```diff
@@ -212,39 +197,43 @@ jobs:
...
+    runs-on: ${{ matrix.runner }}
     strategy:
+      fail-fast: false
       matrix:
-        platform:
-          - runner: macos-12
-            target: x86_64
-          # - runner: macos-14
-          #   target: aarch64
+        runner: [macos-latest]
```
leoromanovsky, line 204: `macos-12` is deprecated

.github/workflows/python.yml
```diff
@@ -212,39 +197,43 @@ jobs:
...
 
   macos:
-    runs-on: ${{ matrix.platform.runner }}
+    runs-on: ${{ matrix.runner }}
     strategy:
+      fail-fast: false
       matrix:
-        platform:
-          - runner: macos-12
-            target: x86_64
```
leoromanovsky, line 220: ironically i can't get `x86_64` to work now but `aarch64` does

.github/workflows/python.yml
rasendubi: > only testing against latest python, increasing the matrix size seems like over-kill

that's a bad idea because there are important differences between versions and we also had compatibility issues in the past because we didn't test

leoromanovsky: I'm not against it and tried it in this PR with the 5 supported versions. It did slow down test running but I can certainly see the utility.

rasendubi: ok, so the version on this line does not actually affect what versions we build/test. Actual images are built inside docker images

rasendubi: From build log:
```
🐍 Found CPython 3.8, CPython 3.9, CPython 3.10, CPython 3.11, CPython 3.12, PyPy 3.8, PyPy 3.9, PyPy 3.10
```

.github/workflows/python.yml
rasendubi: What's the error?

rasendubi:
Looks like the new images added python-3.13 and we fail to build for that. To support it, it looks like we need to upgrade pyo3 to 0.23+ but we're held back by serde-pyobject which requires pyo3-0.22. I have submitted a PR here: https://github.com/Jij-Inc/serde-pyobject/pull/13 and here's a fix for our repo: https://github.com/Eppo-exp/eppo-multiplatform/pull/79

leoromanovsky:
> Looks like the new images added python-3.13 and we fail to build for that. To support it, it looks like we need to upgrade pyo3 to 0.23+ but we're held back by serde-pyobject which requires pyo3-0.22. I have submitted a PR here: [Jij-Inc/serde-pyobject#13](https://github.com/Jij-Inc/serde-pyobject/pull/13) and here's a fix for our repo: #79

Rebase with your latest changes ✔️ 

```
Run ls -l dist
total 13672
-rw-r--r--  1 runner  staff  2335283 Nov 27 [17](https://github.com/Eppo-exp/eppo-multiplatform/actions/runs/12054913678/job/33614056238?pr=77#step:5:18):01 eppo_server_sdk-4.1.1-cp311-cp311-macosx_10_12_x86_64.whl
-rw-r--r--  1 runner  staff  2329177 Nov 27 17:02 eppo_server_sdk-4.1.1-cp312-cp312-macosx_10_12_x86_64.whl
-rw-r--r--  1 runner  staff  2328976 Nov 27 17:03 eppo_server_sdk-4.1.1-cp313-cp313-macosx_10_12_x86_64.whl
```

rasendubi (approved):

PR #78, opened 2024-11-26
feat: Add test status badge to README

typotter:


README.md
```diff
@@ -1,5 +1,10 @@
 ## Eppo Multiplatform: SDKs and Artifacts to support Flagging and Experimentation
 
+[![Test](https://github.com/Eppo-exp/eppo-multiplatform/actions/workflows/ci.yml/badge.svg)](https://github.com/Eppo-exp/eppo-multiplatform/actions/workflows/ci.yml)
```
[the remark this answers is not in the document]
typotter: Thanks for doing the work to split them out!

Made the change to the text here, but from what I can gather, it's actually pulling the name from the yaml file. I've updated the name in the yaml file, but since the badge is pulling the status from `main`, it will still say "Test" until we merge here.

.github/workflows/ci.yml
```diff
@@ -1,4 +1,4 @@
-name: Test
+name: Rust SDK
```
rasendubi, line 1: This also handles testing the core and (currently) fastly, so it's not "rust sdk"

rasendubi: I see the discussion in the other thread. I think it's OK to merge but we should address this later

PR #79, opened 2024-11-27
chore(python): support free-threaded cpython-3.13, drop support for python-3.8

rasendubi:
Free-threaded cpython-3.13 requires pyo3-0.23.2.

serde-pyobject has not been updated yet, so I have submitted a [PR to update it](https://github.com/Jij-Inc/serde-pyobject/pull/13) + pinned my patched version for now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,3 +11,6 @@ members = [
 [patch.crates-io]
 # Local override for development.
 eppo_core = { path = './eppo_core' }
+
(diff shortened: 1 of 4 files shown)


rasendubi:
>   error: the configured PyPy interpreter version (3.8) is lower than PyO3's minimum supported version (3.9)

3.8 has actually reached EoL on 2024-10-07 so I think it should be fine to drop.

PR #73, opened 2024-11-22
Quality: Parameterize SDK and test-data branches in CI workflow

typotter:
🎟️ Fixes FF-3141
🎟️ Fixes FF-3095
 towards FF-3085

👯‍♂️ **Related PRs**
- [sdk-test-data#81](https://github.com/Eppo-exp/sdk-test-data/pull/81)

### Motivation
Changes are often made to the [sdk-test-data](https://github.com/Eppo-exp/sdk-test-data) repository to capture new behaviours, bugs and edge cases. When these changes are pushed to `main`, the SDKs are cloned locally (locally to the github action running) and their respective tests are run. These tests are set up and run by copies of the SDK test workflows - see [sdk-test-data workflow](https://github.com/Eppo-exp/sdk-test-data/blob/main/.github/workflows/test-sdks.yml). There are a number of limitations to this setup:

- Test steps are copied from the SDK’s respective workflows; changes to the SDK test workflows need to be replicated in sdk-test-data and this is not obvious to devs
- The SDK tests are not able to run against in-flight changes to `sdk-test-data`
- When new test-data is committed that breaks an SDK, that breakage is not surfaced in the SDK’s repository until some action triggers its testing workflow (on demand, on pull-request, etc.).

### Description of Changes
_This change_
🚀 - the CI  workflow is enhanced into a [reusable workflow](https://docs.github.com/en/actions/sharing-automations/reusing-workflows), exposing parameters for SDK branch and the sdk-test-data branch to use in testing.
- The test workflow runs using the main branch of sdk-test-data on all main pushes and Pull Requests using the PR's branch

_External to this Change_
[sdk-test-data](https://github.com/Eppo-exp/sdk-test-data/blob/main/.github/workflows) get two testing workflows.
1. ♻️  "Local Testing"- For all pull request changes, the "Local Testing" workflow calls the reusable SDK workflows, test results are recorded only in the sdk-test-data repository. This is run using the main SDK branch and the "current" branch of workflow, i.e. the pull request branch. (SDK repo does not see/is not notified of test failures during PR lifecycle, only on push to main)
2. ♻️ 🧑‍💻 "Remote Testing" - On all pushes to the main branch of sdk-test-data, the SDK testing workflows are triggered to run within their respective repositories, alerting all subscribers of failed runs (repo is red/green).

PR #82, opened 2024-11-28
chore(ci): allow calling workflow for testing

rasendubi:
@typotter I would look for an approach like this for #73

Basically:
- override submodule checkout in ci workflow
- don't touch `sdk-test-data` in the rest of workflows

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,17 @@ on:
     branches:
       - main
   pull_request:
+  workflow_dispatch:
+  workflow_call:
+    inputs:
+      sdk_branch:
+        type: string
+        description: The branch of the SDK to test
+        required: false
+      test_data_branch:
+        type: string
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -20,7 +31,17 @@ jobs:
...
+          repository: Eppo-exp/eppo-multiplatform
+          ref: ${{ inputs.sdk_branch || github.ref }}
           submodules: true
+
+      - uses: actions/checkout@v4
+        if: ${{ inputs.test_data_branch }}
+        with:
+          repository: Eppo-exp/sdk-test-data
+          ref: ${{ inputs.test_data_branch }}
+          path: sdk-test-data
```
rasendubi, line 43: I hope that this should be enough but maybe we need to override `submodules` to `false` if `test_data_branch` is supplied

PR #103, opened 2024-12-03
Revert "chore(python): support free-threaded cpython-3.13, drop support for python-3.8"

rasendubi:
Reverts Eppo-exp/eppo-multiplatform#79

I just realized we cannot override dependencies in the workspace because these overrides won't be preserved when we release the crate and will fail to build for our users.

We have to wait till serde-pyobject PR is merged. If that takes too long, we could fork or re-implement our serialization


leoromanovsky (approved):
Good observation, thank you.

PR #104, opened 2024-12-04
Test python client shutdown (FF-3648)

leoromanovsky:


python-sdk/tests/test_client.py
```diff
@@ -50,3 +52,56 @@ def test_get_bandit_keys_some():
...
+
+
+def test_shutdown_clean():
+    client = init("ufc", wait_for_init=True)
+    # Ensure client is fully initialized
+    assert client.is_initialized() == True
+    # Test clean shutdown
+    client.shutdown()
+    # Verify client is no longer initialized
+    assert client.is_initialized() == False
```
rasendubi, line 64: That's a bad assertion. Shutting down does not mean the client is no longer initialized

This goes for all `is_initialized()` assertions below

python-sdk/tests/test_client.py
```diff
@@ -50,3 +52,56 @@ def test_get_bandit_keys_some():
...
+
+
+def test_shutdown_during_operations():
```
rasendubi, line 67: Given how client is structured (poller is not connected to configuration store), this test has zero chances to fail

PR #106, opened 2024-12-04
chore: bump versions

rasendubi:
Would like to push a bunch of releases

### DIFF
--- a/eppo_core/Cargo.toml
+++ b/eppo_core/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "eppo_core"
-version = "4.1.1"
+version = "4.2.0"
 edition = "2021"
 description = "Eppo SDK core library"
 repository = "https://github.com/Eppo-exp/rust-sdk"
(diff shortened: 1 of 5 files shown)


eppo_core/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "eppo_core"
-version = "4.1.1"
+version = "4.2.0"
```
leoromanovsky, line 3: Main change here was removing `reqwest::blocking` and introducing `tokio` for async handling. Do you think this helps/hurts any threading stability? Any additional improvements for handling exceptions in tokio?

leoromanovsky (approved):

eppo_core/Cargo.toml
rasendubi: This should make no difference as we're just doing what `reqwest::blocking` did under the hood

rasendubi: sorry, just found that we included breaking changes in 4.1.1, so we need a major bump: https://github.com/Eppo-exp/eppo-multiplatform/pull/107

PR #109, opened 2024-12-05
chore(ci): fix rust publish workflow

rasendubi:
Rust publish workflow has [failed](https://github.com/Eppo-exp/eppo-multiplatform/actions/runs/12166755031/job/33933721997) because it couldn’t build fastly-edge-assignments.

Split fastly-edge-assignments off workspace, so it’s not build on release.


leoromanovsky (approved):
Cool nice separation - we will remove the server completely from this repo as soon as we can extract the precomputed logic into `eppo_core`

rasendubi:
yep, I moved edge stuff to the end of workflow file, so we can easily drop it later. I'm going to defer eppo_core@5.0.0 release till we get #81 / #110 in as it might ship another breaking change to the core

PR #110, opened 2024-12-05
refactor(core): move PrecomputedConfiguration into core

rasendubi:
@leoromanovsky this is opened to merge into #81

### DIFF
--- a/eppo_core/src/eval/eval_precomputed_assignments.rs
+++ b/eppo_core/src/eval/eval_precomputed_assignments.rs
@@ -2,29 +2,58 @@ use std::collections::HashMap;
 use std::sync::Arc;
 
 use chrono::{DateTime, Utc};
+use serde::{Deserialize, Serialize};
 
 use crate::eval::get_assignment;
-use crate::ufc::Assignment;
-use crate::ufc::{AssignmentFormat, Environment};
-use crate::{Attributes, Configuration, EvaluationError, Str};
+use crate::ufc::{Assignment, AssignmentFormat, Environment, ValueWire, VariationType};
+use crate::{Attributes, Configuration, Str};
 
-#[derive(Debug)]
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
 pub struct PrecomputedConfiguration {
-    pub created_at: DateTime<Utc>,
-    pub format: AssignmentFormat,
-    pub environment: Environment,
-    pub flags: HashMap<String, Result<Assignment, EvaluationError>>,
+    created_at: DateTime<Utc>,
+    /// `format` is always `AssignmentFormat::Precomputed`.
+    format: AssignmentFormat,
+    // Environment might be missing if configuration was absent during evaluation.
+    environment: Option<Environment>,
+    flags: HashMap<String, PrecomputedAssignment>,
 }
 
-impl PrecomputedConfiguration {
-    pub fn empty(environment_name: &str) -> Self {
-        Self {
-            created_at: Utc::now(),
-            format: AssignmentFormat::Precomputed,
-            environment: Environment {
-                name: environment_name.into(),
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct PrecomputedAssignment {
+    variation_type: VariationType,
+    variation_value: ValueWire,
+
+    do_log: bool,
+    // If `do_log` is false, the client doesn’t need the field below.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    allocation_key: Option<Str>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    variation_key: Option<Str>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    extra_logging: Option<HashMap<String, String>>,
+}
+
+impl From<Assignment> for PrecomputedAssignment {
+    fn from(assignment: Assignment) -> PrecomputedAssignment {
+        match assignment.event {
+            Some(event) => PrecomputedAssignment {
+                variation_type: assignment.value.variation_type(),
+                variation_value: assignment.value.variation_value(),
+                do_log: true,
+                allocation_key: Some(event.base.allocation.clone()),
+                variation_key: Some(event.base.variation.clone()),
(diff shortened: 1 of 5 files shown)


eppo_core/src/eval/eval_precomputed_assignments.rs
```diff
@@ -2,29 +2,58 @@ use std::collections::HashMap;
...
-        Self {
-            created_at: Utc::now(),
-            format: AssignmentFormat::Precomputed,
-            environment: Environment {
-                name: environment_name.into(),
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PrecomputedAssignment {
+    variation_type: VariationType,
+    variation_value: ValueWire,
```
rasendubi, line 26: One diff here is that JSON assignment are now encoded as strings — following UFC flags config approach

eppo_core/src/eval/eval_precomputed_assignments.rs
```diff
@@ -2,29 +2,58 @@ use std::collections::HashMap;
...
+
+    do_log: bool,
+    // If `do_log` is false, the client doesn’t need the field below.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    allocation_key: Option<Str>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    variation_key: Option<Str>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    extra_logging: Option<HashMap<String, String>>,
```
rasendubi, line 35: Decided to skip serializing these if `do_log` is false for now. There is no reason for the client to need these except the logging purposes

eppo_core/src/eval/eval_precomputed_assignments.rs
```diff
@@ -2,29 +2,58 @@ use std::collections::HashMap;
...
 pub struct PrecomputedConfiguration {
-    pub created_at: DateTime<Utc>,
-    pub format: AssignmentFormat,
-    pub environment: Environment,
-    pub flags: HashMap<String, Result<Assignment, EvaluationError>>,
+    created_at: DateTime<Utc>,
+    /// `format` is always `AssignmentFormat::Precomputed`.
+    format: AssignmentFormat,
+    // Environment might be missing if configuration was absent during evaluation.
+    environment: Option<Environment>,
```
rasendubi, line 18: There's no good default environment value, so decided to `null` it instead of sending random string.

eppo_core/src/ufc/assignment.rs
```diff
@@ -29,10 +31,38 @@ pub enum AssignmentValue {
     /// A boolean value.
     Boolean(bool),
     /// Arbitrary JSON value.
-    Json(Arc<serde_json::Value>),
+    Json {
+        raw: Str,
```
rasendubi, line 35: Saving raw json value here. This will also allow us to return raw string in some SDKs if users want to use their own parser.

eppo_core/src/eval/eval_precomputed_assignments.rs
```diff
@@ -2,29 +2,58 @@ use std::collections::HashMap;
...
 
-#[derive(Debug)]
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
 pub struct PrecomputedConfiguration {
-    pub created_at: DateTime<Utc>,
-    pub format: AssignmentFormat,
-    pub environment: Environment,
-    pub flags: HashMap<String, Result<Assignment, EvaluationError>>,
+    created_at: DateTime<Utc>,
```
rasendubi, line 14: Another change here is that all fields are private to `eppo_core`: edge server shouldn't need to tinker with them and clients would have their own parsers.

Basically, the only way to interact with `PrecomputedConfiguration` is via serialize/deserialize.

This should allow us tweaking the type here without being afraid breaking the edge server.

eppo_core/src/eval/evaluator.rs
```diff
@@ -117,14 +117,12 @@ impl Evaluator {
...
         subject_key: &Str,
         subject_attributes: &Arc<Attributes>,
-        early_exit: bool,
     ) -> PrecomputedConfiguration {
         let configuration = self.get_configuration();
         get_precomputed_assignments(
             configuration.as_ref().map(AsRef::as_ref),
             &subject_key,
             &subject_attributes,
-            early_exit,
```
leoromanovsky, line 127: 👍 ok your philosophy is not to implement something without a use, I can understand that

eppo_core/src/ufc/assignment.rs
```diff
@@ -29,10 +31,38 @@ pub enum AssignmentValue {
...
+        parsed: Arc<serde_json::Value>,
+    },
+}
+
+impl PartialEq for AssignmentValue {
```
leoromanovsky, line 40: cool!

leoromanovsky (commented):
Just a single question on `do_log` case

eppo_core/src/eval/eval_precomputed_assignments.rs
```diff
@@ -2,29 +2,58 @@ use std::collections::HashMap;
...
+                variation_key: Some(event.base.variation.clone()),
+                extra_logging: Some(event.base.extra_logging.clone()),
+            },
+            None => PrecomputedAssignment {
+                variation_type: assignment.value.variation_type(),
+                variation_value: assignment.value.variation_value(),
+                do_log: true,
+                allocation_key: None,
+                variation_key: None,
+                extra_logging: None,
```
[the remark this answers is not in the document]
sameerank: Yup, `doLog` in the js client sdk is used to control if the assignment is logged, and if there's no allocation match, we shouldn't invoke the logger callback (if that's how the `do_log` check is also getting used in eppo_core)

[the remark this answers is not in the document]
rasendubi: Good catch — that was a typo. `do_log` should be `false` here

eppo_core/src/ufc/assignment.rs
rasendubi: haha, `PartialEq` was auto-generated before but with adding `raw` json strings, we can't compare on them (because serialization may differ), so I had to roll my own implementation — that's rather un-cool 😅

leoromanovsky (approved):

PR #112, opened 2024-12-06
chore: Add Dockerfile and instructions for building ruby gem locally

felipecsl:


ruby-sdk/ext/eppo_client/Cargo.toml
```diff
@@ -12,7 +12,7 @@ crate-type = ["cdylib"]
 
 [dependencies]
 env_logger = { version = "0.11.3", features = ["unstable-kv"] }
-eppo_core = { version = "=5.0.0", features = ["vendored"] }
+eppo_core = { version = "=4.1.1", features = ["vendored"] }
```
rasendubi, line 15: If we're building from local source, it must depend on the latest (possibly unpublished) core. Otherwise, it's undefined what you're building — it's not the latest version and it's the previous version.

felipecsl: @rasendubi I agree. However, I had to change this because the build was failing since it couldn't find `eppo_core` v5.0.0 - possibly it was trying to download it instead of using the local one.
How do I force it to build it from local instead?

felipecsl: thank you!

build.sh
```diff
@@ -0,0 +1,16 @@
...
+export RUSTUP_HOME=$RUST_INSTALL_PATH/rustup
+export CARGO_HOME=$RUST_INSTALL_PATH/cargo
+export PATH=$CARGO_HOME/bin:$PATH
+export RB_SYS_DOCK_CACHE_DIR=$PWD/cache
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+rustup install stable
+rustup default stable
+mkdir -p /.cargo
+echo "[patch.crates-io.eppo_core]" > "/.cargo/config.toml"
+echo "path = '$PWD/eppo_core'" >> "/.cargo/config.toml"
```
felipecsl, line 13: @rasendubi I added the patch here but this doesn't seem to be working with `rb-sys-dock`, do you have any bandwidth to take a look? that would be highly appreciated

rasendubi: oh, that might be because `rb-sys-dock` runs another docker container for the actual build, so it does not copy cargo config. There should be a parameter for `rb-sys-dock` to run a command before the build

felipecsl:
Alright finally got time to wrap up this PR. I decided to simplify and incorporate some feedback from @yfrancis. By instructing users to clone a specific ruby-sdk tag, we don't have to worry about building the gem against a local version of eppo-core. Also this makes much more sense than trying to build directly from `main` which is undesirable.
PTAL @rasendubi @leoromanovsky

build.sh
```diff
@@ -0,0 +1,16 @@
...
+cargo --config /.cargo/config.toml fetch
+cd ruby-sdk
+rb-sys-dock --platform <platform> --mount-toolchains --ruby-versions 3.3 --build -V
```
felipecsl, line 16: we could probably find a way to parameterize this by having an `ARG PLATFORM` in the Dockerfile and passing that as a parameter to this script, but I didn't want to mess with that quite yet. Future improvement

ruby-sdk/README.md
```diff
@@ -43,3 +43,18 @@ Make sure you remove the override before updating `Cargo.lock`. Otherwise, the l
 * Bump versions in `ruby-sdk/lib/eppo_client/version.rb` and `ruby-sdk/ext/eppo_client/Cargo.toml`
 * Run `cargo update --workspace --verbose` to update `Cargo.lock`
 * Run `bundle` to update `Gemfile.lock`
+
+
+## Building Ruby native lib
```
leoromanovsky, line 48: 👍

Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM ruby:3.3
```
leoromanovsky, line 1: This is a top-level Dockerfile but only deals w/ ruby.

* move it and `build.sh` to the `ruby-sdk` or
* change it to `Dockerfile.ruby.build` and `build-ruby.sh`

leoromanovsky (approved):

PR #113, opened 2024-12-06
FF-3674 feat: obfuscate precomputed configuration

rasendubi:


eppo_core/src/precomputed.rs
```diff
@@ -0,0 +1,180 @@
...
+    variation_value: Base64Str,
+    do_log: bool,
+    // If `do_log` is false, the client doesn’t need the fields below.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    allocation_key: Option<Base64Str>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    variation_key: Option<Base64Str>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    extra_logging: Option<HashMap<Base64Str, Base64Str>>,
+}
```
rasendubi, line 90: @typotter @sameerank This is the format for obfuscated configuration that edge function is going to produce. Hope it's easy to see where md5 (`Md5HashedStr`) or base64 (`Base64Str`) is used.

All md5 hashes are computed by prepending `ObfuscatedPrecomputedConfiguration::salt` to the string. It's a 16 random bytes encoded as base64 in the config. We need to base64-decode it before hashing. Note that because it's random bytes and not a string, [`btoa` may fail](https://developer.mozilla.org/en-US/docs/Glossary/Base64#javascript_support) and we need to use array version of decoding in js

fastly-edge-assignments/src/handlers/handler_assignments.rs
```diff
@@ -129,7 +129,7 @@ pub fn handle_assignments(mut req: Request) -> Result<Response, Error> {
 
     // Create an HTTP OK response with the assignments
     Response::from_status(StatusCode::OK)
-        .with_body_json(&precomputed_configuration)
+        .with_body_json(&precomputed_configuration.obfuscate())
```
rasendubi, line 132: kinda proud that edge service change is 1 word 😎 

cc @leoromanovsky

eppo_core/src/precomputed.rs
rasendubi: Here's an example for flags in sdk-test-data / flags-v1.json:
<details>

```json
{
  "obfuscated": true,
  "format": "PRECOMPUTED",
  "salt": "/Ls8/hc2x1wDOKQ33syRnw==",
  "created_at": "2024-12-06T17:36:11.116639Z",
  "environment": {
    "name": "Test"
  },
  "flags": {
    "ecf30449f115fa08881ed49d4124749a": {
      "variation_type": "NUMERIC",
      "variation_value": "My4xNDE1OTI2",
      "do_log": true,
      "allocation_key": "cm9sbG91dA==",
      "variation_key": "cGk=",
      "extra_logging": {}
    },
    "3bab431fe39dc3cdd6831b3c6ec1e6a3": {
      "variation_type": "STRING",
      "variation_value": "b2xk",
      "do_log": true,
      "allocation_key": "bnVsbC1vcGVyYXRvcg==",
      "variation_key": "b2xk",
      "extra_logging": {}
    },
    "550bb44d14f53e13eb6a1f1082c20c0f": {
      "variation_type": "JSON",
      "variation_value": "eyAiaW50ZWdlciI6IDIsICJzdHJpbmciOiAidHdvIiwgImZsb2F0IjogMi4wIH0=",
      "do_log": true,
      "allocation_key": "NTAvNTAgc3BsaXQ=",
      "variation_key": "dHdv",
      "extra_logging": {}
    },
    "ec055680f8e0b41f47b70cf15a8c175f": {
      "variation_type": "INTEGER",
      "variation_value": "MQ==",
      "do_log": true,
      "allocation_key": "NTAvNTAgc3BsaXQ=",
      "variation_key": "b25l",
      "extra_logging": {}
    },
    "24bcd66520102821e592040a24960ec0": {
      "variation_type": "BOOLEAN",
      "variation_value": "ZmFsc2U=",
      "do_log": true,
      "allocation_key": "b2ZmLWZvci1hbGw=",
      "variation_key": "b2Zm",
      "extra_logging": {}
    },
    "b0a851a2072ee5278ced2680c22e4627": {
      "variation_type": "STRING",
      "variation_value": "Y3VycmVudA==",
      "do_log": true,
      "allocation_key": "Y3VycmVudC12ZXJzaW9ucw==",
      "variation_key": "Y3VycmVudA==",
      "extra_logging": {}
    }
  }
}
```

</details>

fastly-edge-assignments/src/handlers/handler_assignments.rs
typotter: won't this break existing clients (which don't support obfuscated?)

eppo_core/src/timestamp.rs
```diff
@@ -0,0 +1,8 @@
+//! A simple wrapper around `chrono`, so users don’t have to depend on it directly.
```
typotter, line 1: 💪

fastly-edge-assignments/src/handlers/handler_assignments.rs
leoromanovsky: That's the best kind of change 😄 

FYI the JS browser client does not currently support obfuscation decoding and is deployed in production in our FE product on `eppo.cloud`. 

I am thinking whether we should do a conditional obfuscation, but that raises the question of why we would ever want "not obfuscated". Since we still don't have any public customers for this I would argue we could:

* Keep the code as-is
* Modify the JS-client to only accept obfuscated configuration; reject currently published version as not workable.

rasendubi: I would say that it's too early to treat precomputed configs as "in production." There's no real users, rolled out implementation doesn't follow the spec, and it's already broken by fixes in https://github.com/Eppo-exp/eppo-multiplatform/pull/110. I was a bit surprised to learn that JS SDK support was already released given that implementation is so off the spec

rasendubi: > but that raises the question of why we would ever want "not obfuscated".

"Evaluating" non-obfuscated configuration is significantly faster. That might be one reason

sameerank: Yes we SHOULD modify the js-client to only accept obfuscated precomputed flags. I agree with that direction!

Some options for moving forward:
1. Stop dogfooding in the Eppo app (switch back to the normal config-evaluation-based client) so we can make this change and deprecate the current js-client-sdk version
2. Version the edge compute URL so that obfuscated and unobfuscated precomputed flags have different endpoints, and release a new js-client-sdk version that points to the new endpoint

sameerank: > "Evaluating" non-obfuscated configuration is significantly faster. That might be one reason

What would be the bottleneck? Base64 decoding?

rasendubi: @sameerank there's a couple of changes in #110 that break unobfuscated format as well, so there will still be some breakage in #2:
- logging fields (`allocation_key`, `variation_key`, `extra_logging`) are now optional. Previously, the service was skipping allocations with `doLog: false`
- JSON values are strings (same as UFC). Previously they were shipped as-is

sameerank: Let's go with option 1 then. We'll pause dogfooding and start again when the js-client-sdk is updated to use obfuscated precomputed assignments

aarsilv: I agree we don't need to support unobfuscated just for our dogfooding. Pausing and restarting makes sense.

One use case for non-obfuscated responses is for live subject evaluation. So I think we still want the edge function to be able to do it, such as when we hit it with a JWT rather than SDK key

However, no _SDK_ needs to be able to request and process a non-obfuscated response I don't think.

leoromanovsky: Ok we can create a separate handler for the JWT auth and it can serve not-obfuscated if needed. Lets do this after we move it to the private repo please.

eppo_core/src/precomputed.rs
sameerank: A few thoughts from looking at the example (thanks for putting that together!):

1. Looks like I need to add `variationType` to the decoded items in https://github.com/Eppo-exp/js-sdk-common/blob/main/src/decoding.ts#L81-L88
2. We've been using camelCase for the UFC and so far also for the precomputed object, e.g. https://github.com/Eppo-exp/js-sdk-common/blob/d994780e0916520700bea5947085f38f5d0c677a/src/client/eppo-precomputed-client.spec.ts#L18-L66. Not generally opposed to snake_case, but it would be inconsistent in this situation
3. Could be worth having some key-values in `extraLogging` to illustrate `decodeObject` in action
4. I also need to update https://github.com/Eppo-exp/js-sdk-common/blob/d994780e0916520700bea5947085f38f5d0c677a/src/client/eppo-precomputed-client.ts#L278-L280 to use the `salt` when hashing the flag key

rasendubi: Good catch about `camelCase` — it should be `camelCase`, I just forgot the annotation.

rasendubi: For `extraLogging`, both keys and values are base64-encoded:
```python
{
  ...,
  "extraLogging": {
    "aGVsbG8K": "d29ybGQK",
  }
}
```

eppo_core/src/obfuscation.rs
```diff
@@ -0,0 +1,69 @@
...
+use crate::{ufc::ValueWire, Str};
+
+/// `md5::Digest` that implements `Serialize` and `Deserialize` (by converting to hex-encoded
+/// string).
+#[serde_as]
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub(crate) struct Md5HashedStr(#[serde_as(as = "serde_with::hex::Hex")] [u8; 16]);
+
+impl Md5HashedStr {
+    pub fn new(salt: &[u8], input: &[u8]) -> Md5HashedStr {
```
aarsilv, line 13: `salt` 💪

eppo_core/src/precomputed.rs
```diff
@@ -0,0 +1,180 @@
...
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct PrecomputedAssignment {
+    pub(crate) variation_type: VariationType,
+    pub(crate) variation_value: ValueWire,
+
+    pub(crate) do_log: bool,
+    // If `do_log` is false, the client doesn’t need the field below.
```
aarsilv, line 31: Nice thinking!

eppo_core/src/precomputed.rs
```diff
@@ -0,0 +1,180 @@
...
+        };
+
+        let obfuscated = configuration.obfuscate();
+        let flag_key = Md5HashedStr::new(&obfuscated.salt, b"test-flag");
+        let flag = obfuscated.flags.get(&flag_key);
+
+        assert!(flag.is_some());
+        assert_eq!(
+            serde_json::to_string(flag.unwrap()).unwrap(),
+            r#"{"variation_type":"STRING","variation_value":"aGVsbG8sIHdvcmxkIQ==","do_log":true,"allocation_key":"YWxsb2NhdGlvbi1rZXk=","variation_key":"dmFyaWF0aW9uLWtleQ==","extra_logging":{"aGVsbG8=":"d29ybGQ="}}"#
```
aarsilv, line 177: 👌

aarsilv (approved):
I don't speak rust, but from what I was able to gather things are looking good here--nice work!! 🎉

eppo_core/src/precomputed.rs
sameerank: I've written out how this will look within the "configuration wire" format here (as I'm understanding it): https://github.com/Eppo-exp/sdk-test-data/pull/89/files

sameerank: I think the one question that came to mind in the process is if this is the right level to nest the `obfuscated` boolean. Will `subjectKey` and `subjectAttributes` (which are in the parent object) ever be in an obfuscated format and can we use the child object's attribute to determine if they are obfuscated?

eppo_core/src/ufc/models.rs
```diff
@@ -34,11 +34,11 @@ pub(crate) struct UniversalFlagConfigWire {
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
-#[serde(rename_all = "UPPERCASE")]
-pub enum AssignmentFormat {
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
```
leoromanovsky, line 37: `SCREAMING_SNAKE_CASE` is my favorite part of rust so far 😃

eppo_core/src/obfuscation.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn serialize_md5() {
+        let md5_digest = md5::compute(b"hello");
+        let s = Md5HashedStr::from(md5_digest);
```
leoromanovsky, line 63: doesn't hurt to do booleans and numbers here too

eppo_core/src/obfuscation.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// Convert value to base64. Booleans and numbers are converted to string first.
+    fn from(value: ValueWire) -> Base64Str {
+        let s = match value {
+            ValueWire::Boolean(b) => Str::from_static_str(if b { "true" } else { "false" }),
+            ValueWire::Number(n) => n.to_string().into(),
+            ValueWire::String(s) => s,
+        };
+        Base64Str(s)
+    }
+}
```
leoromanovsky, line 54: just want to say that i am finding this rust pattern of 10 implementing desired traits and 2) pattern matching, very elegant. I was a little worried that rust would be like c++ and golang and not particularly "fun" to write but so far it's my top 3 favorite language.

leoromanovsky (approved):
rust code looks great. please wait for a review from @sameerank or @typotter when they are satisfied that the client/server coordination is done.

eppo_core/src/precomputed.rs
```diff
@@ -0,0 +1,180 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PrecomputedConfiguration {
+    pub(crate) obfuscated: serde_bool::False,
+    pub(crate) created_at: Timestamp,
+    /// `format` is always `AssignmentFormat::Precomputed`.
+    pub(crate) format: ConfigurationFormat,
+    // Environment might be missing if configuration was absent during evaluation.
+    pub(crate) environment: Option<Environment>,
+    pub(crate) flags: HashMap<String, PrecomputedAssignment>,
```
typotter, line 21: in the SDK-SDK response, we'll include subject and subject attributes here, where there is the `obfuscated` param. cc @sameerank who does bring up a good point about these fields. Should they be obfuscated?

typotter (approved):

sameerank (approved):
Approving because it matches up with the specs in the design doc. I think it's fine to iterate on this in a subsequent PR to obfuscate the subject key and attributes too. Also will further work be done in `eppo_core` to make it match the "configuration wire" format? I attempted to put together an example of it here: https://github.com/Eppo-exp/sdk-test-data/pull/89/files#diff-66ef842bf0a570eabadbd6cb4dbd22ef931180c8d27ba223f1f019176e562a15 Or is that done somewhere else?

PR #115, opened 2024-12-09
bug: wrap poller thread in `std::panic::catch_unwind` (FF-3468)

leoromanovsky:
## Motivation

Customer observed segfaults sometimes during process shutdown.

## Description

**Change 1 (Likely Root Cause):**
When the parent process disconnects from the thread’s shared channel, the thread attempts another run. During shutdown, this may cause the thread to access memory that the Python interpreter has already released.

**Change 2:**
Wraps the thread’s execution in a catch_unwind (try/catch equivalent) to handle unexpected panics more gracefully. (edited)

Screenshot 2024-12-09 at 10 20 32 AM

**Change 3: Already merged**

There is an unreleased improvement to the `poller_thread` that handles exceptions from HTTP errors and returns a `Result` - https://github.com/Eppo-exp/eppo-multiplatform/commit/41376295f59ba9f6325c1abe83ebb4f147785d43

The customers' environment did not have this change and it could also be the root cause.

## Deployment

The `eppo_core@5.0.0` branch is not yet released - it was our intention to do so after some upcoming features are merged (https://github.com/Eppo-exp/eppo-multiplatform/pull/113).

1. To expedite this improvement to the poller I can merge and deploy python `4.1.3` against `eppo_core@4.x.y`
2. We can merge the PR above and roll both our with the version numbers in this PR.

### DIFF
--- a/eppo_core/src/poller_thread.rs
+++ b/eppo_core/src/poller_thread.rs
@@ -114,19 +114,20 @@ impl PollerThread {
     ) -> std::io::Result<PollerThread> {
         // Using `sync_channel` here as it makes `stop_sender` `Sync` (shareable between
         // threads). Buffer size of 1 should be enough for our use case as we're sending a stop
-        // command and we can simply `try_send()` and ignore if the buffer is full (another thread
-        // has send a stop command already).
+        // command, and we can simply `try_send()` and ignore if the buffer is full (another thread
+        // has sent a stop command already).
         let (stop_sender, stop_receiver) = std::sync::mpsc::sync_channel::<()>(1);
 
         let result = Arc::new((Mutex::new(None), Condvar::new()));
 
         let join_handle = {
+            // Cloning Arc for move into thread
             let result = Arc::clone(&result);
             let update_result = move |value| {
                 *result.0.lock().unwrap() = Some(value);
                 result.1.notify_all();
             };
-        
+
             std::thread::Builder::new()
                 .name("eppo-poller".to_owned())
                 .spawn(move || {
@@ -141,7 +142,7 @@ impl PollerThread {
                                 return;
                             }
                         };
-        
+
                         loop {
                             log::debug!(target: "eppo", "fetching new configuration");
                             let result = runtime.block_on(fetcher.fetch_configuration());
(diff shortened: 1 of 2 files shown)


eppo_core/src/poller_thread.rs
```diff
@@ -121,61 +121,68 @@ impl PollerThread {
...
+                                }
                             }
                         }
+                    }));
+        
+                    // If catch_unwind returns Err, it means a panic occurred.
+                    if let Err(_panic_info) = result {
+                        // Handle the panic gracefully by updating the result with an error.
+                        update_result(Err(Error::PollerThreadPanicked));
                     }
```
leoromanovsky, line 186: new handler

eppo_core/src/poller_thread.rs
```diff
@@ -121,61 +121,68 @@ impl PollerThread {
...
-                        let result = runtime.block_on(fetcher.fetch_configuration());
-                        match result {
-                            Ok(configuration) => {
-                                store.set_configuration(Arc::new(configuration));
-                                update_result(Ok(()))
-                            }
-                            Err(err @ (Error::Unauthorized | Error::InvalidBaseUrl(_))) => {
-                                // Unrecoverable errors
-                                update_result(Err(err));
+                    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
```
leoromanovsky, line 133: The change is wrapping in `std::panic::catch_unwind`

eppo_core/src/poller_thread.rs
```diff
@@ -121,61 +121,68 @@ impl PollerThread {
...
+                                    std::thread::sleep(timeout);
+                                }
                             }
                         }
+                    }));
+        
+                    // If catch_unwind returns Err, it means a panic occurred.
+                    if let Err(_panic_info) = result {
+                        // Handle the panic gracefully by updating the result with an error.
+                        update_result(Err(Error::PollerThreadPanicked));
```
aarsilv, line 185: 👍

aarsilv (approved):
I don't speak rust but the catching of the panic and then gracefully handling sounds good to me and matches the inclination of GPT!

eppo_core/src/poller_thread.rs
```diff
@@ -131,51 +131,61 @@ impl PollerThread {
...
+                                    log::debug!(target: "eppo", "poller thread received stop command");
+                                    // Stop command received, break out of the loop to end the thread.
+                                    return;
+                                }
+                                Err(RecvTimeoutError::Disconnected) => {
+                                    // When the other end of channel disconnects, calls to
+                                    // .recv_timeout() return immediately.
+                                    // Stop the thread.
+                                    log::debug!(target: "eppo", "poller thread received disconnected");
+                                    return;
```
leoromanovsky, line 179: this was modified from `std::thread::sleep` to returning

PR #123, opened 2024-12-16
fix(rust,ruby): fix serialization format for AttributeValue

rasendubi:
There's an unreleased bug in Ruby whereas get_json_assignment returned two-field object with :raw and :parsed fields instead of just parsed value.

This is due to AttributeValue changing internal format for JSON variant.

Implement Serialize manually for AttributeValue, so only serialize parsed value.

This commit also adds implementation for magnus' IntoValue/TryConvert in eppo core, so we're slightly less reliant on magnus_serde.

### DIFF
--- a/eppo_core/src/ufc/assignment.rs
+++ b/eppo_core/src/ufc/assignment.rs
@@ -1,6 +1,7 @@
 use std::sync::Arc;
 
-use serde::{Deserialize, Serialize};
+use serde::ser::SerializeStruct;
+use serde::Serialize;
 
 use crate::{events::AssignmentEvent, Str};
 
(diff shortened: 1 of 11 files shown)


eppo_core/src/ufc/assignment.rs
```diff
@@ -37,6 +48,38 @@ pub enum AssignmentValue {
...
+                state.serialize_field("value", b)?;
+            }
+            AssignmentValue::Json { raw: _, parsed } => {
+                state.serialize_field("type", "JSON")?;
+                state.serialize_field("value", parsed)?;
+            }
+        }
+        state.end()
+    }
+}
```
rasendubi, line 81: This is the real fix. The rest is just moving away from magnus_serde.

I initially wanted to remove `Serialize` implementation completely and rely on Ruby-specific conversion functions. But that took quite a bit of work and we still need magnus_serde to translate Json's parsed value to Ruby, and it broke Rust's serialization of evaluation details.

So I just bit the bullet and implemented `Serialize` for `AssignmentValue`.

package.json
```diff
@@ -4,7 +4,7 @@
     "test": "npm run with-server 'npm-run-all test:*'",
     "test:rust": "cargo test",
     "test:python": "cd python-sdk && pytest",
-    "test:ruby": "cd ruby-sdk && bundle exec rake test",
+    "test:ruby": "cd ruby-sdk && bundle exec rake build && bundle exec rspec",
```
rasendubi, line 7: This is the fix for CI: `bundle exec rake test` does not run rspec

leoromanovsky: Good find - thank you for adding this.

PR #128, opened 2024-12-17
feat(core): precomputed bandits

rasendubi:


eppo_core/src/eval/eval_precomputed.rs
```diff
@@ -0,0 +1,326 @@
...
+                                    Some((variation_value.clone(), precomputed_bandit))
+                                })
+                                .collect()
+                        };
+
+                    Some((flag_key.clone(), flag_bandits))
+                })
+                .collect()
+        })
+        .unwrap_or_default();
```
rasendubi, line 160: The diff in this file is a bit large because I renamed the file.

This highlighted piece is the main change, though

eppo_core/src/precomputed.rs
rasendubi, line 1: This is another important file to review

eppo_core/src/precomputed.rs
```diff
@@ -99,23 +130,42 @@ impl PrecomputedConfiguration {
 
 impl From<PrecomputedConfiguration> for ObfuscatedPrecomputedConfiguration {
     fn from(config: PrecomputedConfiguration) -> Self {
-        let salt = rand::thread_rng().gen();
+        let salt: Str = {
+            let bytes = rand::thread_rng().gen::<[u8; 16]>();
+            base64::prelude::BASE64_STANDARD_NO_PAD
+                .encode(&bytes)
+                .into()
+        };
```
rasendubi, line 138: This change is to make it use salt in text mode instead of binary as before

eppo_core/src/str.rs
```diff
@@ -97,6 +97,12 @@ impl std::ops::Deref for Str {
     }
 }
 
+impl std::borrow::Borrow<str> for Str {
+    fn borrow(&self) -> &str {
+        self.as_str()
+    }
+}
```
rasendubi, line 104: This allows us using `&str` for querying `HashMap<Str, _>`.

Quite a bit of changes in this PR is changing `String` -> `Str` to enable quick cloning.

I start leaning to using `Str` whenever we use strings because it's much cheaper to clone and has small string optimization (small strings are stored without heap allocation), so it's a good fit for our use case. These optimizations also make it much faster for converting to/from other languages (e.g., Python). The only downside is that occasional `String -> Str` conversion might be a bit more costly but there are few

.github/workflows/ci.yml
```diff
@@ -49,9 +49,3 @@ jobs:
 
       - run: cargo test --verbose --workspace
       - run: cargo doc --verbose
-
-      # Add WASM target
-      - run: rustup target add wasm32-wasi
-      # Build WASM target separately
-      - run: cargo build --verbose --target wasm32-wasi
-        working-directory: fastly-edge-assignments
```
leoromanovsky, line 57: thanks for tidying this up

eppo_core/src/eval/evaluator.rs
```diff
@@ -76,7 +76,7 @@ impl Evaluator {
         flag_key: &str,
         subject_key: &Str,
         subject_attributes: &ContextAttributes,
-        actions: &HashMap<String, ContextAttributes>,
+        actions: &HashMap<Str, ContextAttributes>,
```
leoromanovsky, line 79: thanks for the consistent use of `Str`

eppo_core/src/precomputed.rs
leoromanovsky: cc @typotter

eppo_core/src/eval/eval_precomputed.rs
```diff
@@ -0,0 +1,326 @@
...
+                .compiled
+                .flags
+                .iter()
+                .filter_map(|(flag_key, flag)| {
+                    let flag = flag.as_ref().ok()?;
+
+                    // Skip non-string variations as they can't be bandits.
+                    if flag.variation_type != VariationType::String {
+                        return None;
+                    }
```
leoromanovsky, line 77: I know this is true but this optimization could put us into a tricky spot if the code ships in clients and this fact changes. cc @aarsilv is this ever changing?

```suggestion

```

eppo_core/src/eval/eval_precomputed.rs
leoromanovsky: Thanks for the note; yep I recognize the existing stuff 😄

eppo_core/src/precomputed.rs
```diff
@@ -99,23 +130,42 @@ impl PrecomputedConfiguration {
...
                 .into_iter()
                 .map(|(k, v)| {
                     (
-                        Md5HashedStr::new(&salt, k.as_bytes()),
+                        Md5HashedStr::new(salt.as_bytes(), k.as_bytes()),
                         ObfuscatedPrecomputedAssignment::from(v),
                     )
                 })
                 .collect(),
+            bandits: config
```
leoromanovsky, line 154: do you have the bandit obfuscation as part of a unit test?

leoromanovsky (approved):

eppo_core/src/precomputed.rs
rasendubi: There is a small test to compute bandits against `bandit-flags-v1.json`+`bandit-models-v1.json` and I verified the output.

Overall, I'm more reliant on types here rather than tests. Having obfuscated type clearly define where md5s and base64s go makes it trivial to ensure correctness. (If the type requires an md5, there's only one way to get — hash salt+string — regular strings and base64 don't fit.)

Once we stabilize formats, I do want to add a couple of test cases to sdk-test-data though

PR #130, opened 2024-12-17
per-flag bandit action filters (FF-3734)

leoromanovsky:


Cargo.toml
```diff
@@ -7,10 +7,6 @@ members = [
     "ruby-sdk/ext/eppo_client",
 ]
 
-exclude = [
-  "fastly-edge-assignments",
-]
-
```
leoromanovsky, line 13: this code is completed deleted now - my IDE was complaining about it here

eppo_core/src/eval/eval_precomputed.rs
```diff
@@ -76,6 +76,18 @@ pub fn get_precomputed_configuration(
...
+                    let flag_actions = if let Some(actions) = actions.get(flag_key) {
+                        actions
+                    } else {
+                        log::warn!(
+                            target: "eppo",
+                            subject_key,
+                            flag_key;
+                            "No actions provided for flag that requires them"
+                        );
+                        return None;
```
leoromanovsky, line 88: The JS commons implementation of `getAllBandits` (https://github.com/Eppo-exp/js-sdk-common/pull/180/files) produces this outcome by iterating over the `actions` keys - if a flag key does not have an actions key then it will not receive a bandit evaluation. This achieves the same thing.

rasendubi (approved):

PR #136, opened 2024-12-31
FF-3363 chore: streamline version strings management

rasendubi:
Our current versioning flow is tricky and error-prone. With multiple packages in the same repository and changes that affect multiple SDKs, it's hard to keep track what SDKs need a version bump and what kind of version bump they need (major/minor/patch). Keeping track of required version bumps is taxing, and already led to multiple incidents when versions were not properly bumped (breaking changes released without major bump, or version was accidentally skipped).

This PR tries to make version string management easier by integrating [changesets](https://github.com/changesets/changesets).

### New flow

The new flow will look like this:
- When we add new changes, we should add changeset files (`npm run changeset` helps to generate it automatically) that describe what packages need what kinds of version bumps, along with the description of the changes (there are examples of these in this PR).
  - There's a github bot that's going to remind us to add changeset files (optionally because not all changes should cause a release). See changeset-bot below
- Once merged, a github workflow opens a PR that applies all version bumps and updates changelogs. If there are multiple unreleased versions, the workflow will update the existing PR.

  Version updating is two-step (internally):
  - `changesets` updates versions in `package.json`s
  - A bunch of small scripts update `Cargo.toml`s and keep all version strings in sync (e.g., updating `eppo_core` dependency version in all SDKs, or syncing version strings in Ruby's Cargo.toml and version.rb files). This is [recommended by changesets](https://github.com/changesets/changesets/blob/main/docs/versioning-apps.md).
  
- When we are ready to release, we merge the PR and create github releases as usual.

### How this was tested

This PR includes changesets for all unreleased changes. I have applied them with `npm run version` and verified the results.

### Why not nx

When discussing this before, someone proposed nx. Theoretically, it's a good suggestion as it can do more and works with different languages (not just js). Practically, I couldn't integrate it because their Rust plugin doesn't work for mixed-language projects (which is exactly what our SDKs are: python bridging to rust, ruby bridging to rust).

If that is fixed, we can still migrate to nx as nx uses the same format for changeset files.


rasendubi:
Ruby failure is unrelated to this PR. Fix: https://github.com/Eppo-exp/eppo-multiplatform/pull/137

rasendubi:
Yes, I have seen semantic-release. My main concern is that we don't follow semantic commits strictly, changes are often mislabeled, breaking changes are not properly tagged sometimes (in the PR that introduces them), squashing often erases the history (when multiple kinds of changes are introduced in the same PR), and our commit messages are messy and not really suitable as changelogs. With semantic-release, the only way to fix that is rewriting the git history and force-pushing to `main`, which I would like to avoid.

Changesets, on the other hand, decouples this information from our git history, so it's really easy manage/edit/fix it (e.g., if we find that the change was incorrectly tagged (or not) as major change) and retrofitting it (like I did in this PR).

README.md
```diff
@@ -26,19 +26,17 @@ Eppo is a modular flagging and experimentation analysis tool. Eppo's SDKs are bu
 
 ### Release process
 
-To release a new version of SDK:
-1. Make sure that version strings have been updated:
-   - Eppo core: [eppo_core/Cargo.toml](eppo_core/Cargo.toml)
-   - Rust: [rust-sdk/Cargo.toml](rust-sdk/Cargo.toml)
-   - Python: [python-sdk/Cargo.toml](python-sdk/Cargo.toml)
-   - Ruby: [ruby-sdk/lib/eppo_client/version.rb](ruby-sdk/lib/eppo_client/version.rb) and [ruby-sdk/ext/eppo_client/Cargo.toml](ruby-sdk/ext/eppo_client/Cargo.toml)
+To release new versions:
```
leoromanovsky, line 29: Just watched this video https://www.youtube.com/watch?v=puQYAhpfpkA to get a sense for how this would fit into our workflow, very interesting! I understand what you are saying on the PR comment that this doesn't require developers to be very strict w/ commits and we can just "be good once" at the release step.

I am open to trying this tool out for a few months.

I also understand @felipecsl 's feedback about better commit messages - that is absolutely something we should be as well, separate from the tool we use. Would you suggest a linter rule to enforce/remind this?

.changeset/cuddly-houses-share.md
```diff
@@ -0,0 +1,8 @@
+---
```
leoromanovsky, line 1: From the video I watched - my understanding is that when we are ready to public a new release, in that PR these files will be deleted and the versions bumped.

leoromanovsky (approved):

.changeset/cuddly-houses-share.md
rasendubi: Yes. I'm going to merge this PR now and the bot should open a bump PR in a minute

rasendubi: small fix: https://github.com/Eppo-exp/eppo-multiplatform/pull/140 🙂

rasendubi: here: https://github.com/Eppo-exp/eppo-multiplatform/pull/141

PR #137, opened 2024-12-31
chore(ci): properly quote ruby versions

rasendubi:
YAML interpreted 3.0 as a number and stripped ".0" when installing Ruby. So this caused the latest Ruby version to be installed (instead of 3.0).

With Ruby-3.4 recently released, this broke our CI because ruby-3.4 is not currently supported by our version of rb_sys.


felipecsl (approved):
nice catch, thanks!

PR #142, opened 2025-01-07
fix(ruby): Fix some minor Ruby style issues

felipecsl:
* `return` keyword is optional and can be omitted for the last expression in a method
* proper use of early return guards
* misc cleanup

### DIFF
--- a/ruby-sdk/lib/eppo_client/client.rb
+++ b/ruby-sdk/lib/eppo_client/client.rb
@@ -22,7 +22,7 @@ module EppoClient
     def init(config)
       config.validate
 
-      if !@core.nil?
+      if @core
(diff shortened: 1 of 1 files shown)


leoromanovsky (approved):
personally i don't like `unless`, as someone who doesn't read a lot of ruby code it takes me extra time to grok it but if you feel its canonical ruby that's cool

felipecsl:
> personally i don't like `unless`, as someone who doesn't read a lot of ruby code it takes me extra time to grok it but if you feel its canonical ruby that's cool

Yeah it's good practice but only for simple conditionals, eg: `unless foo || bar.nil?` is not recommended and one should use an `if` instead

PR #143, opened 2025-01-07
chore(events): Implement event delivery to ingestion endpoint

felipecsl:
## Description

Follow up from #132 - allow evnets to be delivered to an ingestion URL

## TODOs

Non-comprehensive list of pending work items:

* Handling events with failed ingestion
* Retrying
* Persistence

### DIFF
--- a/eppo_core/src/events/event_delivery.rs
+++ b/eppo_core/src/events/event_delivery.rs
@@ -1,0 +1,60 @@
+use crate::events::event::Event;
+use crate::Error;
+use log::info;
+use reqwest::{Method, StatusCode};
+use serde::{Deserialize, Serialize};
+
+pub struct EventDelivery {
+    sdk_key: String,
+    ingestion_url: String,
+    client: reqwest::Client,
+}
+
+#[derive(serde::Deserialize)]
+struct EventDeliveryResponse {
+    failed_events: Vec<String>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct IngestionRequestBody {
+    eppo_events: Vec<Event>,
+}
+
+/// Responsible for delivering event batches to the Eppo ingestion service.
+impl EventDelivery {
+    pub fn new(sdk_key: String, ingestion_url: String) -> Self {
+        let client = reqwest::Client::new();
+        EventDelivery {
+            sdk_key,
+            ingestion_url,
+            client,
+        }
+    }
+
+    // Delivers the provided event batch and returns a Vec with the events that failed to be delivered.
+    pub async fn deliver(self, events: &[Event]) -> Result<Vec<String>, Error> {
+        let ingestion_url = self.ingestion_url;
+        let sdk_key = self.sdk_key;
+        info!("Delivering {} events to {}", events.len(), ingestion_url);
+        let body = IngestionRequestBody { eppo_events: events.to_vec() };
+        let serialized_body = serde_json::to_string(&body).expect("Failed to serialize body");
+        let response = self.client.request(Method::POST, ingestion_url)
+            .header("Content-Type", "application/json")
+            .header("X-Eppo-Token", sdk_key)
+            .body(serialized_body)
+            .send()
+            .await?;
+        let response = response.error_for_status().map_err(|err| {
+            return if err.status() == Some(StatusCode::UNAUTHORIZED) {
+                log::warn!(target: "eppo", "client is not authorized. Check your API key");
+                Error::Unauthorized
+            } else {
+                log::warn!(target: "eppo", "received non-200 response while fetching new configuration: {:?}", err);
+                Error::from(err)
+            }
+        })?;
+        let response = response.json::<EventDeliveryResponse>().await?;
+        info!("Batch delivered successfully, {} events failed ingestion", response.failed_events.len());
+        Ok(response.failed_events)
+    }
+}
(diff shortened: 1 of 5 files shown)


eppo_core/src/events/event_delivery.rs
```diff
@@ -0,0 +1,60 @@
...
+        let sdk_key = self.sdk_key;
+        info!("Delivering {} events to {}", events.len(), ingestion_url);
+        let body = IngestionRequestBody { eppo_events: events.to_vec() };
+        let serialized_body = serde_json::to_string(&body).expect("Failed to serialize body");
+        let response = self.client.request(Method::POST, ingestion_url)
+            .header("Content-Type", "application/json")
+            .header("X-Eppo-Token", sdk_key)
+            .body(serialized_body)
+            .send()
+            .await?;
```
rasendubi, line 46: nit: reqwest has `.post()` and `.json()` shortcuts:
```suggestion
        let response = self.client.post(ingestion_url)
            .header("X-Eppo-Token", sdk_key)
            .json(&body)
            .send()
            .await?;
```

eppo_core/src/events/event_delivery.rs
```diff
@@ -0,0 +1,60 @@
...
+        let client = reqwest::Client::new();
+        EventDelivery {
+            sdk_key,
+            ingestion_url,
+            client,
+        }
+    }
+
+    // Delivers the provided event batch and returns a Vec with the events that failed to be delivered.
+    pub async fn deliver(self, events: &[Event]) -> Result<Vec<String>, Error> {
```
rasendubi, line 35: minor: `EventDeliveryResponse` is more specific than a generic `Vec<String>`. (When I see `Vec<String>`, I have no idea what it is.)
```suggestion
    pub async fn deliver(self, events: &[Event]) -> Result<EventDeliveryResponse, Error> {
```

eppo_core/src/events/event_dispatcher.rs
```diff
@@ -51,7 +54,10 @@ impl<'a> EventDispatcher {
         let batch_size = config.batch_size;
         loop {
             let mut batch_queue: Vec<Event> = Vec::with_capacity(batch_size);
-            let ingestion_url = config.ingestion_url.clone();
+            let event_delivery = EventDelivery::new(
+                config.sdk_key.clone(),
+                config.ingestion_url.clone()
+            );
```
rasendubi, line 60: minor(perf): this creates a new reqwest client on each batch. Reqwest recommends against that because the client manages some state (e.g., concurrent requests and reusing an active connection to the server)

eppo_core/src/events/event_delivery.rs
```diff
@@ -0,0 +1,60 @@
...
+
+pub struct EventDelivery {
+    sdk_key: String,
+    ingestion_url: String,
+    client: reqwest::Client,
+}
+
+#[derive(serde::Deserialize)]
+struct EventDeliveryResponse {
+    failed_events: Vec<String>,
```
rasendubi, line 15: I believe we want `Uuid` here, so we can match it back to events we just sent out. (As a bonus, `Uuid` is more efficient than `String` for copying and comparison.)
```suggestion
    failed_events: Vec<uuid::Uuid>,
```

felipecsl:
> I would like to suggest drafting a quick and dirty design doc with the implementation approach that we envision, so we can map out tricky parts together and have a smooth sailing afterwards. Let me know if you need help with writing the doc or want to jump on a call to discuss it

That's a great idea. I'll ping you on Slack!

eppo_core/src/events/event_dispatcher.rs
felipecsl: image
i couldn't figure out how to reuse without running into this borrow checker error

eppo_core/src/events/event_delivery.rs
```diff
@@ -0,0 +1,62 @@
...
+        let response = self.client.post(ingestion_url)
+            .header("X-Eppo-Token", sdk_key)
+            .json(&body)
+            .send()
+            .await?;
+        let response = response.error_for_status().map_err(|err| {
+            return if err.status() == Some(StatusCode::UNAUTHORIZED) {
+                // TODO: Mark all events as failed, indicate that this error is not-retriable
+                log::warn!(target: "eppo", "client is not authorized. Check your API key");
+                Error::Unauthorized
```
rasendubi, line 51: good catch about unauthorized!

minor: For configuration polling, that usually means that SDK key is not valid, so we just stop the poller. Does it makes sense to do the same here? If we failed to send one batch because we're unauthorized, sending the next batch would likely fail in the same way

rasendubi (approved):

eppo_core/src/events/event_delivery.rs
felipecsl: yeah exactly the same!

eppo_core/src/events/event_delivery.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct IngestionRequestBody {
+    eppo_events: Vec<Event>,
+}
+
+/// Responsible for delivering event batches to the Eppo ingestion service.
+impl EventDelivery {
+    pub fn new(sdk_key: String, ingestion_url: String) -> Result<Self, ParseError> {
```
rasendubi, line 28: minor: in rust when taking ownership, it's preferable to ask for exact types that you need (e.g., `Str` and `Url` here). If the caller has the appropriate types, it can pass the values in as is (instead of potentially converting `Str` to `String`, only to get converted back to `Str`)
```suggestion
    pub fn new(sdk_key: Str, ingestion_url: Url) -> Self {
```

rasendubi: I added it after you merged, so no worries :)

PR #144, opened 2025-01-08
bug: publish macos wheels for python 3.9, 3.10, 3.11 and 3.12 (FF-3786)

leoromanovsky:
## Research:

* https://www.maturin.rs/distribution.html#build-wheels

```
Usage: maturin build [OPTIONS] [ARGS]...

Arguments:
  [ARGS]...
          Rustc flags

Options:
  -i, --interpreter [<INTERPRETER>...]
          The python versions to build wheels for, given as the executables of interpreters such as `python3.9` or `/usr/bin/python3.8`

  -f, --find-interpreter
          Find interpreters from the host machine
```

* https://github.com/pydantic/pydantic-core/blob/main/.github/workflows/ci.yml#L474C1-L484C32

## Testing

We actions log: https://github.com/Eppo-exp/eppo-multiplatform/actions/runs/12680906304/job/35343576093?pr=144

Screenshot 2025-01-08 at 3 39 25 PM
Screenshot 2025-01-08 at 3 39 43 PM

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -220,19 +220,25 @@ jobs:
             target: x86_64
           - runner: macos-14
             target: aarch64
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        exclude:
+          # aarch64 builds only supported from 3.10 and up
+          - platform: { runner: macos-14, target: aarch64 }
+            python-version: "3.9"
     steps:
       - uses: actions/checkout@v4
         with:
           submodules: true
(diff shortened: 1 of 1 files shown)


felipecsl (approved):
🚢

rasendubi (commented):
minor: I wonder if we can install all python versions on the same machine, so we can build all wheels in one action (similar to how we build linux wheels). The majority of time is compiling Rust, which could be reused between python versions

PR #149, opened 2025-01-10
chore(ci): add a workflow to automatically update ruby lockfile

rasendubi:
This new workflow should run automatically after eppo_core releases, update Ruby SDK lockfile (updating eppo_core hash to published one), and open a PR.


leoromanovsky (approved):
Thanks for the automation

PR #154, opened 2025-01-13
feat(events): Add `VecEventQueue` abstraction, event status and attempts

felipecsl:
## Description

* Add `QueuedEvent` with `status` and `attempts`
* Add `EventQueue` trait and simple `VecEventQueue` implementation
* Write a few more tests
* Allow returning a batch of events with a given status
* Enforce max queue size constraint

## Testing
Wrote tests

### DIFF
--- a/eppo_core/src/eval/eval_details_builder.rs
+++ b/eppo_core/src/eval/eval_details_builder.rs
@@ -230,9 +230,10 @@ impl EvalBanditVisitor for EvalDetailsBuilder {
 }
 
 impl<'b> EvalAssignmentVisitor for &'b mut EvalDetailsBuilder {
-    type AllocationVisitor<'a> =
-        <EvalDetailsBuilder as EvalAssignmentVisitor>::AllocationVisitor<'a>
-    where Self: 'a;
+    type AllocationVisitor<'a>
+        = <EvalDetailsBuilder as EvalAssignmentVisitor>::AllocationVisitor<'a>
+    where
+        Self: 'a;
 
     fn visit_allocation<'a>(&'a mut self, allocation: &Allocation) -> Self::AllocationVisitor<'a> {
         EvalAssignmentVisitor::visit_allocation(*self, allocation)
@@ -252,7 +253,8 @@ impl<'b> EvalAssignmentVisitor for &'b mut EvalDetailsBuilder {
 }
 
 impl EvalAssignmentVisitor for EvalDetailsBuilder {
-    type AllocationVisitor<'a> = EvalAllocationDetailsBuilder<'a>
+    type AllocationVisitor<'a>
+        = EvalAllocationDetailsBuilder<'a>
     where
         Self: 'a;
 
@@ -302,11 +304,13 @@ impl EvalAssignmentVisitor for EvalDetailsBuilder {
 }
 
 impl<'b> EvalAllocationVisitor for EvalAllocationDetailsBuilder<'b> {
-    type RuleVisitor<'a> = EvalRuleDetailsBuilder<'a>
+    type RuleVisitor<'a>
+        = EvalRuleDetailsBuilder<'a>
     where
         Self: 'a;
 
-    type SplitVisitor<'a> = EvalSplitDetailsBuilder<'a>
+    type SplitVisitor<'a>
+        = EvalSplitDetailsBuilder<'a>
     where
         Self: 'a;
 
--- a/eppo_core/src/events/event.rs
+++ b/eppo_core/src/events/event.rs
@@ -1,7 +1,7 @@
 use crate::timestamp::Timestamp;
 use serde::{Deserialize, Serialize};
 
-#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
 pub struct Event {
     pub uuid: uuid::Uuid,
     pub timestamp: Timestamp,
--- a/eppo_core/src/events/event.rs
+++ b/eppo_core/src/events/event.rs
@@ -1,7 +1,7 @@
 use crate::timestamp::Timestamp;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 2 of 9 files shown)


eppo_core/src/events/vec_event_queue.rs
```diff
@@ -0,0 +1,197 @@
...
+#[derive(Debug, PartialEq)]
+pub enum QueueError {
+    QueueFull,
+}
+
+// batch size of zero means each event will be delivered individually, thus effectively disabling batching.
+const MIN_BATCH_SIZE: usize = 0;
+const MAX_BATCH_SIZE: usize = 10_000;
+
+impl VecEventQueue {
```
felipecsl, line 34: Basically brought back the initial implementation from #132 and expanded upon it to add max_queue_size and filtering by event status

eppo_core/src/events/vec_event_queue.rs
```diff
@@ -0,0 +1,197 @@
...
+    /// Pushes an event to the end of the queue.
+    fn push(&self, event: QueuedEvent) -> Result<(), QueueError>;
+
+    /// Returns the next batch of events with the provided status.
+    /// and with status PENDING. Events are returned in FIFO order.
+    fn next_batch(&self, status: QueuedEventStatus) -> Vec<QueuedEvent>;
+
+    /// Returns whether the queue contains enough events for delivering *at least* one batch.
+    fn is_batch_full(&self) -> bool;
+}
```
felipecsl, line 15: @rasendubi I'm aware this might conflict with your work, but we can resolve conflicts if needed, I don't expect the trait implementation will diverge too much. What do you think?

eppo_core/src/eval/eval_details_builder.rs
```diff
@@ -302,11 +304,13 @@ impl EvalAssignmentVisitor for EvalDetailsBuilder {
...
 impl<'b> EvalAllocationVisitor for EvalAllocationDetailsBuilder<'b> {
-    type RuleVisitor<'a> = EvalRuleDetailsBuilder<'a>
+    type RuleVisitor<'a>
+        = EvalRuleDetailsBuilder<'a>
     where
         Self: 'a;
 
-    type SplitVisitor<'a> = EvalSplitDetailsBuilder<'a>
+    type SplitVisitor<'a>
+        = EvalSplitDetailsBuilder<'a>
```
felipecsl, line 313: i just ran `cargo fmt` which seems to have modified this file automatically

eppo_core/src/events/event.rs
```diff
@@ -1,7 +1,7 @@
 use crate::timestamp::Timestamp;
 use serde::{Deserialize, Serialize};
 
-#[derive(Debug, Serialize, Deserialize, Clone)]
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
```
rasendubi, line 4: > Clone, PartialEq

minor: when do we need to clone or compare events? Given that `payload` is arbitrarily nested JSON, cloning or comparing it are somewhat slow and we want to avoid that

eppo_core/src/eval/eval_details_builder.rs
rasendubi: Sounds good. Not critical but it would help slightly with reviewing if formatting changes were a separate PR

eppo_core/src/events/event_dispatcher.rs
```diff
@@ -69,15 +75,19 @@ impl EventDispatcher {
                     // Channel closed, no more messages. Exit the main loop.
                     return;
                 }
-                Some(EventDispatcherCommand::Event(event)) => batch_queue.push(event),
+                Some(EventDispatcherCommand::Event(event)) => {
+                    if Err(QueueError::QueueFull) == event_queue.push(event) {
```
rasendubi, line 79: major: we shall move queue pushing from here to `dispatch()` sooner—that'll affect our types, traits, and error handling

eppo_core/src/events/queued_event.rs
```diff
@@ -0,0 +1,47 @@
+use crate::events::event::Event;
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum QueuedEventStatus {
+    PENDING,
+    RETRY,
+    FAILED,
```
rasendubi, line 7: Rust's convention is CamelCase enum variants:
```suggestion
    Pending,
    Retry,
    Failed,
```

eppo_core/src/events/vec_event_queue.rs
rasendubi: It is going to be quite different to account for usage from multiple threads and to avoid re-notification of the delivery task.

I'm in the design phase though, so don't have much code to cause conflict. So let's merge this PR and I will base my work on top

eppo_core/src/events/vec_event_queue.rs
```diff
@@ -0,0 +1,197 @@
...
+        while i < queue.len() && batch.len() < self.batch_size {
+            if queue[i].status == status {
+                batch.push(queue.remove(i).unwrap());
+            } else {
+                i += 1;
+            }
+        }
+
+        batch
+    }
```
rasendubi, line 75: `.remove()` is O(n), so this method is O(n^2). There are ways to make it faster but the simplest is to remove the requirement of filtering—we only ever request pending items, so there's no reason for queue to hold any non-pending items anyway.

rasendubi (approved):

eppo_core/src/events/event.rs
felipecsl: it's just for tests :)

eppo_core/src/events/event_dispatcher.rs
felipecsl: cool, I can take that next

eppo_core/src/events/vec_event_queue.rs
felipecsl: the queue will also hold failed events, so we can retry them. unless you thought about having a separate queue for those? if so then we can get rid of the `status` field

leoromanovsky (approved):
👍

PR #156, opened 2025-01-15
chore(events): Mark events as failed

felipecsl:
## Description

Upon a failed event delivery, mark events `status` as failed, increment the `attempts` count and move them to the `failed` queue.

## Test

* Wrote a test for `mark_events_as_failed`
* Still planning to write another dispatcher test to ensure events are correctly marked as failed. Will do before merging

### DIFF
--- a/eppo_core/src/events/event_dispatcher.rs
+++ b/eppo_core/src/events/event_dispatcher.rs
@@ -27,10 +27,12 @@ pub struct EventDispatcherConfig {
     pub max_retries: Option<u32>,
 }
 
-#[derive(Debug)]
+#[derive(thiserror::Error, Debug)]
 pub enum DispatcherError {
+    #[error("Queue error: {0}")]
     QueueError(QueueError),
-    EventDeliveryError(&'static str),
+    #[error("Receiver should not be closed before all senders are closed")]
+    EventDeliveryError,
 }
 
 /// EventDispatcher is responsible for batching events and delivering them to the ingestion service
@@ -65,9 +67,7 @@ impl<T: EventQueue + Clone + Send + 'static> EventDispatcher<T> {
     pub fn send(&self, command: EventDispatcherCommand) -> Result<(), DispatcherError> {
         match self.tx.send(command) {
             Ok(_) => Ok(()),
-            Err(_) => Err(DispatcherError::EventDeliveryError(
-                "receiver should not be closed before all senders are closed",
-            )),
+            Err(_) => Err(DispatcherError::EventDeliveryError),
         }
     }
 
--- a/eppo_core/src/events/event_dispatcher.rs
+++ b/eppo_core/src/events/event_dispatcher.rs
@@ -5,10 +5,11 @@ use crate::events::vec_event_queue::{EventQueue, QueueError};
 use log::warn;
 use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
 use tokio::time::{Duration, Instant};
+use url::Url;
 
 #[derive(Debug)]
 pub enum EventDispatcherCommand {
-    Event(QueuedEvent),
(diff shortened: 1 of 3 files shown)


eppo_core/src/events/event_dispatcher.rs
```diff
@@ -34,7 +41,7 @@ pub struct EventDispatcher<T> {
     tx: UnboundedSender<EventDispatcherCommand>,
 }
 
-impl<T: EventQueue + Clone> EventDispatcher<T> {
+impl<T: EventQueue + Clone + Send + 'static> EventDispatcher<T> {
```
felipecsl, line 44: unclear if this is good practice or not, but since the event queue is used from the delivery async task, it needs to be `'static`

eppo_core/src/events/event_dispatcher.rs
```diff
@@ -122,28 +108,36 @@ impl<T: EventQueue + Clone> EventDispatcher<T> {
...
+                            let failed_event_uuids = response.failed_events;
+                            if !failed_event_uuids.is_empty() {
+                                warn!("Failed to deliver {} events", failed_event_uuids.len());
+                                let failed_events = batch
+                                    .into_iter()
+                                    .filter(|queued_event| {
+                                        failed_event_uuids.contains(&queued_event.event.uuid)
+                                    })
+                                    .collect();
+                                queue.mark_events_as_failed(failed_events);
```
felipecsl, line 135: this effectively moves the failed events to the failed queue

eppo_core/src/events/event_dispatcher.rs
```diff
@@ -122,28 +108,36 @@ impl<T: EventQueue + Clone> EventDispatcher<T> {
...
-                            if !response.failed_events.is_empty() {
-                                // TODO: Update status, increment attempts and re-enqueue events for retry
-                                warn!("Failed to deliver {} events", response.failed_events.len());
+                            let failed_event_uuids = response.failed_events;
+                            if !failed_event_uuids.is_empty() {
+                                warn!("Failed to deliver {} events", failed_event_uuids.len());
+                                let failed_events = batch
+                                    .into_iter()
+                                    .filter(|queued_event| {
+                                        failed_event_uuids.contains(&queued_event.event.uuid)
```
leoromanovsky, line 132: Are you trying to prevent duplicates if the deleted event is still queued? If so I think that's too heavy handed. What do you think the system behavior looks like without this line.

felipecsl: no, that's not the intention. the goal here is that `response.failed_events` may have a subset of the uuids of events from `batch`, as in, some events in a batch might fail to ingest, so we need to find them and mark as failed for retry. does that make sense?

eppo_core/src/events/event_dispatcher.rs
```diff
@@ -26,6 +27,12 @@ pub struct EventDispatcherConfig {
     pub max_retries: Option<u32>,
 }
 
+#[derive(Debug)]
```
rasendubi, line 30: minor: take a look at `thiserror` (there are examples in eppo core). It helps with deriving boilerplate code, so we can make this type a proper rust error

eppo_core/src/events/event_dispatcher.rs
```diff
@@ -26,6 +27,12 @@ pub struct EventDispatcherConfig {
...
+pub enum DispatcherError {
+    QueueError(QueueError),
+    EventDeliveryError(&'static str),
```
rasendubi, line 33: minor: please prefer named errors + display implementation (see `thiserror`) instead of generic str slices. Strings are hard for the caller to analyze programmatically to decide what to do with the error

rasendubi (approved):

PR #158, opened 2025-01-16
chore(events): Redeliver failed events in batches

felipecsl:
## Description

* Update event loop to attempt delivery of any failed events according to the configured `retry_interval`
* Do not re-enqueue failed events after `max_retries` attempts reached

Note: We're not using `max_retry_delay` or adding any exponential backoff yet. This is a simple loop that just attempts delivery of failed events every `retry_interval` s/ms.

## Testing

Wrote tests


eppo_core/src/events/event_dispatcher.rs
```diff
@@ -78,78 +76,96 @@ impl<T: EventQueue + Clone + Send + 'static> EventDispatcher<T> {
...
+) {
+    if batch.is_empty() {
+        // nothing to deliver
+        return;
+    }
+    // Send events
+    tokio::spawn({
+        async move {
+            // Spawning a new task, so the main task can continue batching events and respond to
+            // commands. At this point, batch is guaranteed to have at least one event.
```
rasendubi, line 135: We have agreed to make delivery "single-threaded" for the first version, so we can just drop `spawn` and do everything inline.

Otherwise, we don't have a concurrency limit, so it may cause unlimited number of concurrent http requests (maybe still limited by reqwest client but probably too high)

eppo_core/src/events/event_dispatcher.rs
```diff
@@ -78,78 +76,96 @@ impl<T: EventQueue + Clone + Send + 'static> EventDispatcher<T> {
...
+            let events = batch
+                .iter()
+                .map(|queued_event| queued_event.clone().event)
```
rasendubi, line 138: Can we do without cloning the whole batch of events? Maybe it's better to change request type, so it does not consume events

eppo_core/src/events/event_dispatcher.rs
```diff
@@ -78,78 +76,96 @@ impl<T: EventQueue + Clone + Send + 'static> EventDispatcher<T> {
...
+            let result = event_delivery.deliver(events).await;
+            match result {
+                Ok(response) => {
+                    let failed_event_uuids = response.failed_events;
+                    if !failed_event_uuids.is_empty() {
+                        warn!("Failed to deliver {} events", failed_event_uuids.len());
+                        let failed_events = batch
+                            .into_iter()
+                            .filter(|queued_event| {
+                                failed_event_uuids.contains(&queued_event.event.uuid)
```
rasendubi, line 149: `failed_event_uuids` is a vector, so `contains()` is O(n). Because this operation is wrapped in another O(n) (`filter`), that totals to O(n^2). It's cheaper to convert `failed_event_uuids` to a hashset and then do the check (I think we can simply change the type in `EventDeliveryResponse`, so it deserializes directly into set)

eppo_core/src/events/event_dispatcher.rs
```diff
@@ -78,78 +76,96 @@ impl<T: EventQueue + Clone + Send + 'static> EventDispatcher<T> {
...
+                            .collect();
+                        queue.mark_events_as_failed(failed_events);
                     }
                 }
-            });
+                Err(err) => {
+                    warn!("Failed to deliver events: {}", err);
+                    // In this case there is no point in retrying delivery since the error is
+                    // non-retriable.
+                }
```
rasendubi, line 159: Timeouts, network errors, and 5xx response fall into this branch and are pretty much retriable

eppo_core/src/events/event_delivery.rs
```diff
@@ -51,12 +51,12 @@ impl EventDelivery {
...
         let response = response.error_for_status().map_err(|err| {
             return if err.status() == Some(StatusCode::UNAUTHORIZED) {
-                // TODO: Mark all events as failed, indicate that this error is not-retriable
+                // This error is not-retriable
                 log::warn!(target: "eppo", "client is not authorized. Check your API key");
                 Error::Unauthorized
             } else {
-                // TODO: Mark all events as failed, indicate that this error is not-retriable
-                log::warn!(target: "eppo", "received non-200 response while fetching new configuration: {:?}", err);
+                // This error is not-retriable
```
rasendubi, line 58: 5xx, 408, and 429 errors are retriable

eppo_core/src/events/vec_event_queue.rs
```diff
@@ -14,15 +16,22 @@ pub trait EventQueue {
...
+}
+
 /// A simple event queue that stores events in a vector
 #[derive(Debug, Clone)]
 pub struct VecEventQueue {
-    batch_size: usize,
-    max_queue_size: usize,
-    event_queue: Arc<Mutex<HashMap<QueuedEventStatus, VecDeque<QueuedEvent>>>>,
+    config: VecEventQueueConfig,
+    event_queue: Arc<Mutex<HashMap<QueuedEventStatus, LinkedHashSet<QueuedEvent>>>>,
```
rasendubi, line 34: nit: hash map on event status is somewhat inefficient given there's just two statuses that we're interested in. Two fields would avoid the hashmap lookup + split the mutex potentially decreasing contention? Pending queue could also continue using `VecDeque`, which is more efficient than `LinkedHashSet`

rasendubi, line 34: minor: `LinkedHashSet` uses a linked list underneath and is not super-efficient for batched operations (or for memory consumption)

I know that you're trying to preserve events order but we already lost it by moving failed events to another queue (or just allowing partial failures to be re-submitted, or submitting multiple batches concurrently). So I don't think it's worth maintaining.

Looking at the code a bit more, we're still popping events from the front and always inserting to the back. I don't think `LinkedHashSet` gives us anything at all.

eppo_core/src/events/event_dispatcher.rs
```diff
@@ -78,78 +76,96 @@ impl<T: EventQueue + Clone + Send + 'static> EventDispatcher<T> {
...
-            if !event_queue.is_batch_full() {
-                let deadline = Instant::now() + config.delivery_interval;
-                // Loop until we have enough events to send or reached deadline.
-                loop {
-                    tokio::select! {
-                        _ = tokio::time::sleep_until(deadline) => {
-                            // reached deadline -> send everything we have
-                            break;
+            let delivery_deadline = Instant::now() + config.delivery_interval;
+            let retry_deadline = Instant::now() + config.retry_interval;
```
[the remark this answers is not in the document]
felipecsl: you're absolutely right, let me rethink this for a bit

eppo_core/src/events/vec_event_queue.rs
felipecsl: we don't want duplicate events, thus the `LinkedHashSet`. I agree on using a map being somewhat pointless, I can split into separate queues

eppo_core/src/events/event_dispatcher.rs
felipecsl: good idea!

eppo_core/src/events/event_dispatcher.rs
```diff
@@ -10,7 +10,6 @@ use url::Url;
 #[derive(Debug)]
 pub enum EventDispatcherCommand {
     Event,
-    Flush,
```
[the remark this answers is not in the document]
rasendubi: Even if we don't expose flush to users, we need flush to handle shutdown correctly.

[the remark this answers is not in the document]
felipecsl: cool, I guess with the new design, this enum is no longer and we have `BatchedMessage` built into `BatchedMessage`, so we should be fine

eppo_core/Cargo.toml
```diff
@@ -38,9 +38,9 @@ serde-bool = "0.1.3"
 serde_json = "1.0.116"
 serde_with = { version = "3.11.0", default-features = false, features = ["base64", "hex", "macros"] }
 thiserror = "2.0.3"
-tokio = { version = "1.34.0", default-features = false, features = ["macros", "sync", "rt", "time"] }
+tokio = { version = "1.34.0", default-features = false, features = ["macros", "sync", "rt", "time", "test-util"] }
```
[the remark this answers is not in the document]
felipecsl: 👍🏽 will update

felipecsl:
@rasendubi i dont want to have this PR grow massively so would love to get it to a mergeable state and keep iterating in follow up PRs. Still gotta add the exponential backoff, but if you could review it as is and provide any feedback on anything you think it's blocking, that'd be helpful. thanks!

ruby-sdk/lib/eppo_client/config.rb
```diff
@@ -8,7 +8,15 @@ module EppoClient
...
-    def initialize(api_key, assignment_logger: AssignmentLogger.new, base_url: EppoClient::Core::DEFAULT_BASE_URL, poll_interval_seconds: EppoClient::Core::DEFAULT_POLL_INTERVAL_SECONDS, poll_jitter_seconds: EppoClient::Core::DEFAULT_POLL_JITTER_SECONDS, initial_configuration: nil, log_level: nil)
+    def initialize(
+      api_key,
+      assignment_logger: AssignmentLogger.new,
+      base_url: EppoClient::Core::DEFAULT_BASE_URL,
+      poll_interval_seconds: EppoClient::Core::DEFAULT_POLL_INTERVAL_SECONDS,
+      poll_jitter_seconds: EppoClient::Core::DEFAULT_POLL_JITTER_SECONDS,
+      initial_configuration: nil,
+      log_level: nil
+    )
```
felipecsl, line 19: just broke args in separate lines here. Still TBD on `initial_configuration` being unused in `initialize`

eppo_core/src/event_ingestion/delivery.rs
```diff
@@ -0,0 +1,108 @@
...
+        let events_to_deliver = batch
+            .clone()
+            .into_iter()
+            .map(|queued_event| queued_event.event)
+            .collect();
+        let result = event_delivery.deliver(events_to_deliver).await;
+        match result {
+            Ok(response) => {
+                let delivery_status_data = collect_delivery_response(batch, response, max_retries);
+                deliver_status(&delivery_status, delivery_status_data).await;
```
rasendubi, line 56: now would be a good time to call `flush()` callbacks to let the caller know that we're done with their delivery and they can shutdown (or whatever they want to do)

eppo_core/src/event_ingestion/event_delivery.rs
```diff
@@ -48,19 +60,27 @@ impl EventDelivery {
...
-            .await?;
+            .await
+            .map_err(EventDeliveryError::RetriableError)?;
         let response = response.error_for_status().map_err(|err| {
             return if err.status() == Some(StatusCode::UNAUTHORIZED) {
-                // TODO: Mark all events as failed, indicate that this error is not-retriable
+                // This error is not-retriable
                 log::warn!(target: "eppo", "client is not authorized. Check your API key");
-                Error::Unauthorized
+                EventDeliveryError::NonRetriableError(err)
```
rasendubi, line 69: minor: I would say that this error is not just non-retriable—it is "fatal." With non-retriable errors, we skip a batch and send the next one. With unauthorized, there's no way to recover and the deliver task should stop sending any new batches exit immediately.

eppo_core/src/event_ingestion/event_delivery.rs
```diff
@@ -48,19 +60,27 @@ impl EventDelivery {
...
+                // Other errors **might be** retriable
+                log::warn!(target: "eppo", "received non-200 response delivering events: {:?}", err);
+                EventDeliveryError::RetriableError(err)
             }
         })?;
-        let response = response.json::<EventDeliveryResponse>().await?;
+        let response = response
+            .json::<EventDeliveryResponse>()
+            .await
+            .map_err(EventDeliveryError::NonRetriableError)?;
```
rasendubi, line 83: 🐛 timeouts and transient network errors fall into here and should be retried. (JSON serialization errors also fall into here and should not be retried.)

eppo_core/src/event_ingestion/event_delivery.rs
```diff
@@ -48,19 +60,27 @@ impl EventDelivery {
...
+                // This error is not-retriable
+                log::warn!(target: "eppo", "received 400 response delivering events: {:?}", err);
+                EventDeliveryError::NonRetriableError(err)
             } else {
-                // TODO: Mark all events as failed, indicate that this error is not-retriable
-                log::warn!(target: "eppo", "received non-200 response while fetching new configuration: {:?}", err);
-                Error::from(err)
+                // Other errors **might be** retriable
+                log::warn!(target: "eppo", "received non-200 response delivering events: {:?}", err);
+                EventDeliveryError::RetriableError(err)
```
rasendubi, line 77: minor: I think we should do a bit more work distinguishing when we should retry or not. 4xx are usually non-retriable except 408 and 429. 5xx are usually retriable. (Marking this minor as retrying is a safe default.)

eppo_core/src/event_ingestion/event_dispatcher.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// Use the sender to dispatch events and the receiver to receive delivery statuses.
+    fn spawn_event_dispatcher(
+        &self,
+    ) -> (Sender<BatchedMessage<QueuedEvent>>, Receiver<FinishedBatch>) {
+        let config = self.config.clone();
+        let ingestion_url = Url::parse(config.ingestion_url.as_str())
+            .expect("Failed to create EventDelivery. invalid ingestion URL");
+        let event_delivery = EventDelivery::new(config.sdk_key.into(), ingestion_url);
+
+        // TODO: Does this make sense for channel size?
```
rasendubi, line 49: Somewhat. I think it's OK for the first inbound channel, so we can buffer up to a batch of events from the user in case background thread is busy. (max_queue_size` seems more fitting though.)

For other channels, that would add extra buffers between all components which we generally don't want. Especially after batcher which compresses multiple events into one message, so `batch_size` channel after batcher would allow buffering up to `batch_size^2` events (`batch_size` messages each carrying `batch_size` events).

I would say that the rest of the channels should be size 1. (0/rendezvous would be preferred but tokio requires at least 1 buffer.)

.github/workflows/ci.yml
```diff
@@ -45,7 +45,7 @@ jobs:
 
       - run: npm ci
       - run: rustup update ${{ matrix.toolchain }} && rustup default ${{ matrix.toolchain }}
-      - run: cargo build --verbose --all-targets --workspace
+      - run: cargo build --verbose --all-targets --workspace --features "event_ingestion"
```
rasendubi, line 48: minor: `--all-features`?

ruby-sdk/lib/eppo_client/config.rb
rasendubi, line 19: nit: I'm planning to work on this bug today, so it would be preferable if formatting changes were its own PR as to not cause additional conflicts

rasendubi (approved):

eppo_core/src/event_ingestion/delivery.rs
```diff
@@ -0,0 +1,108 @@
...
+};
+use crate::event_ingestion::queued_event::QueuedEvent;
+use log::warn;
+use tokio::sync::mpsc;
+
+pub(super) struct QueuedBatch {
+    pub success: Vec<QueuedEvent>,
+    pub failure: Vec<QueuedEvent>,
+    pub retry: Vec<QueuedEvent>,
+}
```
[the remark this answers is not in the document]
felipecsl: in the `retry` fn we're receiving a `QueuedBatch` and using that to differentiate evens that failed (exceeded max retries), succeeded and are pending retry. 
then, from tests we use `FinishedBatch` to assert on the final state of the pipeline. It works like a listener. We could consolidate both `QueuedBatch` and `FinishedBatch` into the same `struct` though, since their have the exact same structure. Will consider that. for the next PR!

eppo_core/src/event_ingestion/delivery.rs
```diff
@@ -0,0 +1,108 @@
...
+
+impl QueuedBatch {
+    fn with_success(success: Vec<QueuedEvent>) -> Self {
```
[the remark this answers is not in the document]
felipecsl: cool, thanks, i'll rename it to match that!

eppo_core/src/event_ingestion/delivery.rs
```diff
@@ -0,0 +1,108 @@
...
+            retry: Vec::new(),
+            failure,
+        }
+    }
+}
+
+pub(super) async fn delivery(
+    mut uplink: mpsc::Receiver<BatchedMessage<QueuedEvent>>,
+    delivery_status: mpsc::Sender<QueuedBatch>,
+    max_retries: u8,
```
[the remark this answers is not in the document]
felipecsl: sure, but I don't see us ever using a `max_retry` value that large in practice. what is the benefit of retrying so many times? My gut feeling is that one should probably give up after so many failures

eppo_core/src/event_ingestion/event_delivery.rs
felipecsl: agreed in general, but do you think a panic is the right approach here?

ruby-sdk/lib/eppo_client/config.rb
felipecsl: reverting!

eppo_core/src/event_ingestion/event_delivery.rs
rasendubi: No, panic is definitely not good. We should add an enum case to `EventDeliveryError`, so that the caller can differentiate between these kinds of errors and do different things (retry batch, drop the batch, stop completely).

felipecsl: agreed, i'll make a note to tackle this in the next PR, thanks!

eppo_core/src/event_ingestion/event_dispatcher.rs
```diff
@@ -0,0 +1,222 @@
...
+            batcher_downlink_rx,
+            delivery_downlink_tx,
+            config.max_retries,
+            event_delivery,
+        ));
+        tokio::spawn(retry::retry(
+            delivery_downlink_rx,
+            batcher_downlink_tx,
+            retry_downlink_tx,
+        ));
```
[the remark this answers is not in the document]
felipecsl: Thanks for expanding your thoughts here @rasendub - interesting discussion cc @petzel 
I agree that the first approach is simpler and it's in fact what we have currently implemented in the JS common SDK (https://github.com/Eppo-exp/js-sdk-common/blob/main/src/events/default-event-dispatcher.ts#L95-L103) with the caveat that we're re-enqueueing failed events after `max_retries` was reached there, which we should probably change to just drop the events in that case, like we're doing here

my concern with that approach is that, if you have a large `max_retries` configured, it could completely halt delivery of new batches until we're done with all retries. with exponential backoffs, that becomes an even bigger problem. I think it's possible that we run into a situation where a single event in a batch failed ingestion (eg. due to malformed/invalid payload in a way that we didn't predict) but other events could still succeed, but I agree that's probably gonna be the less common scenario (most likely a global error, eg. backend outage).

I currently have retries implemented in a separate component here, I'd propose we stick with this implementation for now and reassess in the future, I see this as a minor concern for now.

eppo_core/src/event_ingestion/event_delivery.rs
felipecsl: 👍🏽 i was debating whether or not to tackle this here. took a note for follow up in a future PR

eppo_core/src/event_ingestion/delivery.rs
felipecsl: can you share an example of how you see `flush` being used? im having a hard time understanding your intention there

rasendubi: `flush` is a tool for guaranteeing that we don't loose events in two scenarios:
- P0: graceful shutdown
- P1: non-graceful shutdown

For graceful shutdown (user's app is shutting down and they can wait up to some reasonable timeout), we want to deliver any queued events and delay the shutdown till that's done (so we don't loose events). I originally added oneshot channels as a way to notify the main thread that "we're done, you can continue with shutdown." We need this for P0 as otherwise we're failing at "don't loose events" requirement.

For non-graceful shutdown (we're running in "adversarial" environment and the server may get killed abruptly), the user might want to get delivery guarantee at some predetermined points (usually after handling every request, or even after every event). In this case, we want to give the user a `.flush()` function that returns when all previously-submitted events are flushed. So in that case, oneshots are used to notify the users that "we're done with your events." (We have agreed that this is a P1 as in "next release" but I see this as supporting "don't loose events" in the majority of cloud envs.)

eppo_core/src/event_ingestion/delivery.rs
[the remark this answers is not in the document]
rasendubi: Why retry needs `failure` and `success` events? That is likely going to delay their pruning from persistence

This actually looks like branching: retriable events go to retry, success/failure go to delivery status.
image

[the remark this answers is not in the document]
felipecsl: you're right, it doesn't, since it's just a passthrough for anything other than `retry` events. I can have `delivery` push those out as you suggested 👍🏽

PR #168, opened 2025-01-20
refactor(core)!: split poller thread into background thread and configuration poller

rasendubi:
In preparation for doing more work in the background (event ingestion), we're refactoring poller thread into a more generic background thread / background runtime with configuration poller running on top of it.

This changes API of the core but should be an invisible for SDKs. One change that may be noticeable is that poller thread may exit faster on shutdown (due to using interruptible timer between fetches instead of non-interruptible sleep).


eppo_core/src/background/thread.rs
```diff
@@ -0,0 +1,52 @@
...
+    /// Prefer `graceful_shutdown()` if you have the time to wait.
+    pub fn kill(&self) {
+        self.runtime.stop();
+    }
+
+    /// Command background activities to stop and wait for thread to terminate.
+    pub fn graceful_shutdown(self) {
+        self.runtime.stop();
+
+        let _ = self.join_handle.join();
```
felipecsl, line 50: out of curiosity, why do we need to assign the return value here if it's unused?

eppo_core/src/configuration_poller.rs
```diff
@@ -0,0 +1,190 @@
...
+    loop {
+        log::debug!(target: "eppo", "fetching new configuration");
+
+        match fetcher.fetch_configuration().await {
+            Ok(configuration) => {
+                store.set_configuration(Arc::new(configuration));
+                update_status(Ok(()));
+            }
+            Err(err @ (Error::Unauthorized | Error::InvalidBaseUrl(_))) => {
+                // There errors are not recoverable. Update result and exit the poller.
```
felipecsl, line 137: typo: "These"

felipecsl (commented):
Would love to see more tests for this new async code, especially since I don't fully grasp what's going on here. @leoromanovsky might want to take a look as well

eppo_core/src/background/thread.rs
rasendubi: `let _ =` is the rust's way of saying "ignore the return value." Sometimes that's required to silence a lint warning (when value is marked as `must_use`), other times it's just a way to signal that there's a value we're ignoring.

rasendubi:
> Would love to see more tests for this new async code, especially since I don't fully grasp what's going on here.

Will add a bit more tests and annotate the PR with explanations tomorrow. There ain't much "new" code though — thread and poller are extracted from PollerThread, so you might want to take a look at old code.

"Runtime" is new and is a combination of two concepts from tokio tutorials:
- [Tokio: Bridging with sync code](https://tokio.rs/tokio/topics/bridging). We're getting a hold of runtime's handle to spawn tasks on it (potentially from sync code).
- [Tokio: Graceful Shutdown](https://tokio.rs/tokio/topics/shutdown). Explains [CancellationToken](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html) and [task tracker](https://docs.rs/tokio-util/0.7.13/tokio_util/task/task_tracker/struct.TaskTracker.html) for shutdown.

Also, all of the changed code is executed as part of all SDKs' test suites, so it's end-to-end tested (but I will try adding some unit tests)

eppo_core/src/background/runtime.rs
```diff
@@ -0,0 +1,218 @@
...
+/// When `BackgroundRuntime` is dropped, all background activities are commanded to stop.
+pub struct BackgroundRuntime {
+    tokio_runtime: tokio::runtime::Handle,
+    /// A cancellation token that gets cancelled when runtime needs to exit.
+    cancellation_token: CancellationToken,
+    /// A set of tasks that are required to exit before the tokio runtime can be safely
+    /// stopped. Rust futures are usually safe to drop, so this is normally not needed. But we may
+    /// need this occasionally (e.g., finish writes to disk, etc.)
+    watched_tasks: TaskTracker,
+}
```
rasendubi, line 19: The two main goals for background runtime (and this PR) are:
- **Allow doing different async stuff concurrently.** Today it's only configuration poller but we want to add event ingestion.
- **Handle graceful shutdown for async tasks.** This is currently unused as we don't need any special cleanup requirement. But we'll need this soon when we add persistence — we want to make sure that we fsync'ed database before closing.

Tokio runtime handle is handling the concurrency part. It's a reference to a tokio runtime, so we can spawn multiple tasks on it. The runtime is presumably running *somewhere* — on a background thread in the current setup, but not necessarily. (In wasm, tokio would run without a thread, and for Rust SDK we'll want to allow the user to share their tokio runtime if they have one. Both are "for future".)

For graceful shutdown, we're using a combination of `CancellationToken` and `TaskTracker` — both are explained in [tokio tutorial](https://tokio.rs/tokio/topics/shutdown):
- `CancellationToken` is used to signal to "tracked" tasks that background runtime (Eppo client) is shutting down and they need to exit.
- `TaskTracker` is watching for tracked tasks and is waiting for them to exit.

This setup needs some cooperation from the tasks (they need to watch the cancellation token), so we're distinguishing between "tracked" tasks that do that and require cleanup, and "untracked" tasks that don't have any cleanup requirements. Most of the tasks don't have cleanup, so shouldn't care and just be "untracked".

eppo_core/src/background/runtime.rs
```diff
@@ -0,0 +1,218 @@
...
+    where
+        F: Future + Send + 'static,
+        F::Output: Send + 'static,
+    {
+        self.tokio_runtime.spawn(future)
+    }
+
+    /// Command background activities to stop and exit.
+    pub(crate) fn stop(&self) {
+        self.watched_tasks.close();
```
rasendubi, line 88: This is telling task tracker that we don't expect any new tasks and it should exit when the currently tracked set of tasks exit. (Return from `tracker.wait()` below.)

eppo_core/src/background/runtime.rs
```diff
@@ -0,0 +1,218 @@
...
+        self.cancellation_token.cancel();
```
rasendubi, line 89: This is telling tracked tasks that they should exit.

eppo_core/src/background/runtime.rs
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    /// Wait for all background activities to stop.
+    pub(super) fn wait(&self) -> impl Future {
+        let tracker = self.watched_tasks.clone();
+        async move { tracker.wait().await }
+    }
```
rasendubi, line 96: This is intended for background runtime *runner* (i.e., background thread) to know when we're done and it's OK stop tokio runtime.

eppo_core/src/background/thread.rs
```diff
@@ -0,0 +1,74 @@
+use super::runtime::BackgroundRuntime;
+
+/// An owning handle to a background thread running tokio runtime.
+///
+/// When the handle is dropped, the tokio runtime is commanded to exit and the thread shuts down.
+pub struct BackgroundThread {
+    join_handle: std::thread::JoinHandle<()>,
+    runtime: BackgroundRuntime,
+}
```
rasendubi, line 9: Background runtime does not own the tokio runtime and it needs something else to actually drive tokio runtime. Background thread is the thing that creates tokio runtime and runs it on a native OS thread. (There will be other ways to run tokio in the future.)

eppo_core/src/configuration_poller.rs
```diff
@@ -0,0 +1,188 @@
...
+        ConfigurationPollerConfig {
+            interval: ConfigurationPollerConfig::DEFAULT_POLL_INTERVAL,
+            jitter: ConfigurationPollerConfig::DEFAULT_POLL_JITTER,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct ConfigurationPoller {
+    status: watch::Receiver<Option<Result<(), crate::Error>>>,
```
rasendubi, line 61: We previously used a mutex + condvar to communicate poller status (to implement `.wait_for_configuration()`). Tokio watch channel does the same but is easier to use.

eppo_core/src/configuration_poller.rs
```diff
@@ -0,0 +1,188 @@
...
+    cancellation_token: CancellationToken,
```
rasendubi, line 62: Cancellation token is stored here if we ever want to stop the configuration poller without stopping the whole background runtime.

eppo_core/src/configuration_poller.rs
```diff
@@ -0,0 +1,188 @@
...
+}
+
+impl ConfigurationPoller {
+    pub async fn wait_for_configuration(&self) -> Result<(), crate::Error> {
+        let mut status_rx = self.status.clone();
+        let status = status_rx
+            .wait_for(|status| status.is_some())
```
rasendubi, line 69: `.wait_for()` is the replacement for the mutex + condvar dance

rust-sdk/src/poller.rs
```diff
@@ -80,14 +90,16 @@ impl PollerThread {
     /// # }
     /// ```
     pub fn wait_for_configuration(&self) -> Result<()> {
-        self.0.wait_for_configuration()
+        self.thread
+            .runtime()
+            .block_on(self.poller.wait_for_configuration())
```
rasendubi, line 95: Another nicety of the background runtime is that it allows us running async functions from sync context (see [Handle::block_on](https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html#method.block_on)). This is generally discouraged because it blocks user's thread but sometimes we actually want this blocking (like here).

rasendubi (commented):
@felipecsl I added tests for background runtime and thread

Also see the comments below with more explanations. Let me know if anything else is unclear!

eppo_core/src/background/runtime.rs
felipecsl: > we'll want to allow the user to share their tokio runtime if they have one

In which cases would we allow that and why? I'm assuming that only applies for users using our rust-sdk, right? Otherwise, how would that do that from Ruby or Python, for example?

rasendubi: Yes, that's only relevant to Rust users. If you already use tokio for your server, you usually want to reuse it as that allows for better task sharing

eppo_core/src/background/runtime.rs
```diff
@@ -0,0 +1,218 @@
...
+            .enable_all()
+            .build()
+            .unwrap();
+
+        let background_runtime = BackgroundRuntime::new(tokio_runtime.handle().clone());
+
+        // Without us calling .stop(), .block_on() below would block indefinitely.
+        background_runtime.stop();
+
+        tokio_runtime.block_on(background_runtime.wait());
```
felipecsl, line 129: 👍🏽

eppo_core/src/background/runtime.rs
```diff
@@ -0,0 +1,218 @@
...
+        // background runtime doesn't wait for it to finish.
+        background_runtime.spawn_untracked(async {
+            loop {
+                tokio::time::sleep(Duration::from_secs(1)).await;
+            }
+        });
+
+        background_runtime.stop();
+
+        tokio_runtime.block_on(background_runtime.wait());
```
felipecsl, line 153: does this cause any leaks? I'm assuming no, but since you mentioned above it "never finishes", I was wondering if there were any unwanted side effects of allowing this

eppo_core/src/background/thread.rs
felipecsl: > Background runtime does not own the tokio runtime

Not sure I understand this, I see that `BackgroundRuntime` owns a `tokio_runtime: tokio::runtime::Handle`

eppo_core/src/background/thread.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    #[test]
+    fn test_example_usage() {
+        let background_thread = BackgroundThread::start().unwrap();
+
+        let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
+
+        background_thread.runtime().spawn_untracked(async move {
+            tx.send(true).unwrap();
+        });
```
felipecsl, line 66: thanks, this makes a lot more sense to me now

felipecsl (approved):
Thanks for your inline PR comments and tests, super helpful. A lot of them actually provide very important context that I think should be left as code comments instead. At least to me they seem pretty important :) but I'll leave that up to you. LGTM

eppo_core/src/background/thread.rs
rasendubi: [`Handle`](https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html) is a non-owning reference to a tokio runtime. [`Runtime`](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html) is the actual thing

felipecsl: ah gotcha

eppo_core/src/background/runtime.rs
rasendubi: No. In Rust, futures are *passive* — they don't do anything unless polled/awaited. (This is different from most other languages where futures/promises are *active* and continue doing their thing even if there's noone watching.) When the runtime exits and we destroy tokio runtime, it deallocates all spawned tasks that didn't exit on their own

felipecsl: Interesting! TIL

rasendubi: This is actually the reason why we need an "executor" (e.g., tokio runtime) in Rust. Under the hood, Rust futures compile to state machines that do nothing on their own and need to be polled. So we need a thing that would drive these state machines at appropriate times. That's what tokio runtime does

felipecsl: Suddenly things make a lot more sense 😂

PR #174, opened 2025-01-22
chore(events): Avoid cloning events for delivery

felipecsl:
Addressing prior feedback here https://github.com/Eppo-exp/eppo-multiplatform/pull/159#discussion_r1919882661

### DIFF
--- a/eppo_core/src/event_ingestion/delivery.rs
+++ b/eppo_core/src/event_ingestion/delivery.rs
@@ -67,7 +67,7 @@ pub(super) async fn delivery(
         }
         let events_to_deliver = batch
             .iter()
-            .map(|queued_event| queued_event.event)
+            .map(|queued_event| &queued_event.event)
             .collect();
         let result = event_delivery.deliver(events_to_deliver).await;
         match result {
--- a/eppo_core/src/event_ingestion/delivery.rs
+++ b/eppo_core/src/event_ingestion/delivery.rs
@@ -66,8 +66,7 @@ pub(super) async fn delivery(
             continue;
         }
         let events_to_deliver = batch
-            .clone()
-            .into_iter()
+            .iter()
             .map(|queued_event| queued_event.event)
             .collect();
         let result = event_delivery.deliver(events_to_deliver).await;
--- a/eppo_core/src/event_ingestion/delivery.rs
+++ b/eppo_core/src/event_ingestion/delivery.rs
@@ -5,6 +5,7 @@ use crate::event_ingestion::event_delivery::{
 use crate::event_ingestion::queued_event::QueuedEvent;
 use log::warn;
 use tokio::sync::mpsc;
+use crate::event_ingestion::event::Event;
 
 #[derive(Debug, PartialEq)]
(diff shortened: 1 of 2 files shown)


eppo_core/src/event_ingestion/event_delivery.rs
```diff
@@ -27,9 +27,9 @@ pub(super) enum EventDeliveryError {
     NonRetriableError(reqwest::Error),
 }
 
-#[derive(Debug, Serialize, Deserialize)]
-struct IngestionRequestBody {
-    eppo_events: Vec<Event>,
+#[derive(Debug, Serialize)]
+struct IngestionRequestBody<'a> {
+    eppo_events: Vec<&'a Event>,
```
rasendubi, line 32: This is going half-way for avoiding copy. It does not copy events but it copies the structure of the vector (~80kb with max batch).

Ideally, we want it to take a slice:
```suggestion
    eppo_events: &'a [Event],
```


This is likely complicated by `QueuedEvent` being different from a normal `Event` but I think we can batch `Event`s directly. (My understanding is that `QueuedEvent` doesn't serve any purpose as we're just discarding all the extra fields. `BatchedMessage` was designed to be used with `Event` directly. Retried events and attempts should be scoped to the retry component.)

felipecsl:
@rasendubi looks like `&[&Event]` works, PTAL?

eppo_core/src/event_ingestion/delivery.rs
```diff
@@ -66,11 +67,10 @@ pub(super) async fn delivery(
...
         }
         let events_to_deliver = batch
-            .clone()
-            .into_iter()
-            .map(|queued_event| queued_event.event)
-            .collect();
-        let result = event_delivery.deliver(events_to_deliver).await;
+            .iter()
+            .map(|queued_event| &queued_event.event)
+            .collect::<Vec<&Event>>();
```
rasendubi, line 72: The actual copying is here, not in the `deliver` call below. `events_to_deliver` is a new vector with up to 10,000 references, so `&[&Event]` has the same drawback as `Vec<&Event>`

felipecsl: I had to ask chatgpt here since I was unclear about how to address this 
image
It seems to me like `&[&Event]` is fine since we're avoiding copying the `Event` objects, we're just allocating a vec with their references, which seems fine to me. is this blocking @rasendubi ?

rasendubi: Not blocking.

This PR is a step forward in a sense that it avoids copying Event but as I [mentioned](https://github.com/Eppo-exp/eppo-multiplatform/pull/174#discussion_r1926816065) it still allocates up to 80kb of references alone. Which seems easy to avoid if we use `BatchedMessage<Event>` instead of `BatchedMessage<QueuedEvent>` — the only extra field that is used (`attempts`) is initialized to its default value (so we can optimize it away), and `delivery` component arguably shouldn't know/care about it given we have the `retry` component to take care of these attempts

rasendubi (approved):

PR #176, opened 2025-01-22
chore(events): Enforce max event serialized length

felipecsl:
Fixes FF-3849

## Description

Enforce that each event's serialized JSON string is under 4096 bytes.

## Testing

Wrote tests

### DIFF
--- a/eppo_core/src/event_ingestion/delivery.rs
+++ b/eppo_core/src/event_ingestion/delivery.rs
@@ -99,16 +99,17 @@ fn collect_delivery_response(
     response: EventDeliveryResponse,
     max_retries: u32,
 ) -> QueuedBatch {
-    let failed_event_uuids = response.failed_events;
-    if failed_event_uuids.is_empty() {
+    if response.is_empty() {
         return QueuedBatch::success(batch);
     }
-    warn!("Failed to deliver {} events", failed_event_uuids.len());
+    let failed_retriable_event_uuids = response.retriable_failed_events;
+    let failed_non_retriable_event_uuids = response.non_retriable_failed_events;
+    warn!("Failed to deliver {} events (retriable)", failed_retriable_event_uuids.len());
     let mut success = Vec::new();
     let mut failure = Vec::new();
     let mut retry = Vec::new();
     for queued_event in batch {
-        if failed_event_uuids.contains(&queued_event.event.uuid) {
+        if failed_retriable_event_uuids.contains(&queued_event.event.uuid) {
             if queued_event.attempts < max_retries {
                 // increment failed attempts count and retry
                 retry.push(QueuedEvent::new_from_failed(queued_event));
@@ -116,6 +117,9 @@ fn collect_delivery_response(
                 // max retries reached, mark as failed
                 failure.push(QueuedEvent::new_from_failed(queued_event));
             }
+        } else if failed_non_retriable_event_uuids.contains(&queued_event.event.uuid) {
+            // event may not be retried
+            failure.push(QueuedEvent::new_from_failed(queued_event));
         } else {
             success.push(queued_event);
         }
--- a/eppo_core/src/event_ingestion/delivery.rs
+++ b/eppo_core/src/event_ingestion/delivery.rs
@@ -1,6 +1,6 @@
 use super::BatchedMessage;
 use crate::event_ingestion::event_delivery::{
-    EventDelivery, EventDeliveryError, EventDeliveryResponse,
+    EventDelivery, EventDeliveryError, DeliveryResult,
 };
 use crate::event_ingestion::queued_event::QueuedEvent;
 use log::warn;
@@ -96,14 +96,14 @@ pub(super) async fn delivery(
 
 fn collect_delivery_response(
     batch: Vec<QueuedEvent>,
-    response: EventDeliveryResponse,
+    result: DeliveryResult,
     max_retries: u32,
 ) -> QueuedBatch {
-    if response.is_empty() {
+    if result.is_empty() {
         return QueuedBatch::success(batch);
     }
-    let failed_retriable_event_uuids = response.retriable_failed_events;
-    let failed_non_retriable_event_uuids = response.non_retriable_failed_events;
+    let failed_retriable_event_uuids = result.retriable_failed_events;
+    let failed_non_retriable_event_uuids = result.non_retriable_failed_events;
     warn!("Failed to deliver {} events (retriable)", failed_retriable_event_uuids.len());
     let mut success = Vec::new();
     let mut failure = Vec::new();
--- a/eppo_core/src/event_ingestion/delivery.rs
+++ b/eppo_core/src/event_ingestion/delivery.rs
@@ -80,12 +80,13 @@ pub(super) async fn delivery(
                 match err {
                     EventDeliveryError::RetriableError(_) => {
                         // Retry later
-                        deliver_status(&delivery_status, QueuedBatch::failure(batch)).await;
+                        deliver_status(&delivery_status, QueuedBatch::retry(batch)).await;
                     }
-                    EventDeliveryError::NonRetriableError(_) => {
+                    _ => {
                         warn!("Failed to deliver events: {}", err);
                         // In this case there is no point in retrying delivery since the error is
                         // non-retriable.
+                        deliver_status(&delivery_status, QueuedBatch::failure(batch)).await;
                     }
                 }
             }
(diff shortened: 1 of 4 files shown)


eppo_core/src/event_ingestion/delivery.rs
```diff
@@ -80,12 +80,13 @@ pub(super) async fn delivery(
                 match err {
                     EventDeliveryError::RetriableError(_) => {
                         // Retry later
-                        deliver_status(&delivery_status, QueuedBatch::failure(batch)).await;
+                        deliver_status(&delivery_status, QueuedBatch::retry(batch)).await;
                     }
-                    EventDeliveryError::NonRetriableError(_) => {
+                    _ => {
```
felipecsl, line 85: still need to do more fine grained error handling here, will address in future PR

eppo_core/src/event_ingestion/event_delivery.rs
```diff
@@ -88,3 +98,74 @@ impl EventDelivery {
         Ok(response)
     }
 }
+
+fn ensure_max_event_size(event: &Event) -> Result<(), EventDeliveryError> {
+    let serialized_event = serde_json::to_vec(event)
+        .map_err(|e| JsonSerializationError(e))?;
```
felipecsl, line 104: this is somewhat wasteful since we're serializing each event and throwing away the output, then later we serialize the entire body again, but this makes the code simpler and doesn't seem worth the tradeoff for optimization

eppo_core/src/event_ingestion/delivery.rs
felipecsl: In the case of `EventPayloadTooLargeError`, we could potentially do either:
1. Just throw away just the event(s) that failed and retry the rest of the batch
2. Instead of returning `EventPayloadTooLargeError`, we can instead have `EventDelivery` log an error and discard those offending events but still deliver the rest of the batch. The downside of this approach is that it 

I'm kinda leaning towards option 2 here but the downside is that it doesn't give the user the opportunity to fix it and it may go unnoticed until they check the application logs. thoughts @rasendubi @petzel ?

felipecsl: ended up doing something different altogether

felipecsl: ended up doing something different altogether

eppo_core/src/event_ingestion/event_delivery.rs
```diff
@@ -7,20 +10,48 @@ use std::collections::HashSet;
...
 #[derive(Clone)]
 pub(super) struct EventDelivery {
     sdk_key: Str,
     ingestion_url: Url,
     client: reqwest::Client,
 }
 
-#[derive(serde::Deserialize)]
+#[derive(Debug, PartialEq)]
 pub(super) struct EventDeliveryResponse {
```
rasendubi, line 23: My understanding is that this is no longer a server response but something else

eppo_core/src/event_ingestion/event_delivery.rs
```diff
@@ -7,20 +10,48 @@ use std::collections::HashSet;
...
 }
 
 #[derive(thiserror::Error, Debug)]
 pub(super) enum EventDeliveryError {
+    #[error("Single event payload too large {0} (expected < {max})", max = MAX_EVENT_SERIALIZED_LENGTH)]
+    EventPayloadTooLargeError(usize),
+    #[error("Failed to serialize events to JSON")]
+    JsonSerializationError(serde_json::Error),
+    #[error("Failed to parse JSON response")]
+    JsonDeserializationError(reqwest::Error),
```
rasendubi, line 54: It's weird that this is storing a `reqwest` error and not a `serde_json` error. It may incorrectly clump network errors into this category (and it does — see my next comment)

eppo_core/src/event_ingestion/event_delivery.rs
```diff
@@ -78,13 +126,100 @@ impl EventDelivery {
             }
         })?;
         let response = response
-            .json::<EventDeliveryResponse>()
+            .json::<RawEventDeliveryResponse>()
             .await
-            .map_err(EventDeliveryError::NonRetriableError)?;
+            .map_err(|e| JsonDeserializationError(e))?;
```
rasendubi, line 131: 🐛 This is incorrectly clumping network errors into json deserialization error. In reqwest, response resolves as soon as response headers are received — without waiting for the whole body to arrive (this is to allow streaming responses, etc). However, this means that `response.json()` may return a network error

eppo_core/src/event_ingestion/event_delivery.rs
```diff
@@ -78,13 +126,100 @@ impl EventDelivery {
...
+            non_retriable_failed_events: failed_validation_events,
+        })
+    }
+}
+
+/// Returns whether the provided event's serialized JSON string is over the length limit
+fn ensure_max_event_size(event: &Event) -> Result<bool, EventDeliveryError> {
+    let serialized_event = serde_json::to_vec(event).map_err(|e| JsonSerializationError(e))?;
+    Ok(serialized_event.len() < MAX_EVENT_SERIALIZED_LENGTH)
+}
```
rasendubi, line 147: minor: this is quite wasteful as it serializes the event but discards the work, so we have to go through serialization again

eppo_core/src/event_ingestion/event_delivery.rs
```diff
@@ -7,20 +10,48 @@ use std::collections::HashSet;
 use url::Url;
 use uuid::Uuid;
 
+const MAX_EVENT_SERIALIZED_LENGTH: usize = 4096;
```
rasendubi, line 13: Could you explain how this constant was chosen? Is it a limit imposed by the server or just picked arbitrarily? (would be great to have it in a comment)

rasendubi (approved):
This should work, so OK to merge.

However, I find this approach more complicated than it needs to be. The event travels through multiple layers of processing, batching, and potentially persistence, just to get discarded at the very end (and somewhat disturbing our batch sizes). It would make more sense to validate and discard the input as soon as we receive it.

Another minor'ish concern is double-serialization. (I was also thinking about it in terms of persistence as we would need to deserialize from db's blob storage to Event type, just to serialize it again when sending to ingestion service.) I'm leaning towards making Event contain serialized json bytes, so db operations are just blob store/load with no parsing or serialization. This would make size checks trivial.

```rust
struct Event {
  id: Uuid,
  json_blob: Vec<u8>,
}
```

eppo_core/src/event_ingestion/event_delivery.rs
felipecsl: yup, i can rename for clarity

eppo_core/src/event_ingestion/event_delivery.rs
felipecsl: yup, this is a limit imposed by the backend. its a number we thought was reasonably large enough to hold most events. we don't expect people to start submitting really large event payloads (that's not the intention), and if they do, then we maybe need to help/educate them on how to use it

felipecsl:
> It would make more sense to validate and discard the input as soon as we receive it.

We thought about it, but didn't think it was appropriate to perform JSON serialization up front when the user calls track and potentially slow it down, since we want that call to be fast. It's a tradeoff for sure. We're making an assumption that this won't trigger for 99%+ for events, so it doesn't make sense to pay that cost in 100% of them, if that makes sense.

> Another minor'ish concern is double-serialization

Agree it's not ideal. I think once we have persistence, we can front-load serialization at DB insertion time.

eppo_core/src/event_ingestion/event_delivery.rs
felipecsl: I don't think so, unless I'm reading this wrong
image
we already check for network errors above on line 112

eppo_core/src/event_ingestion/event_delivery.rs
felipecsl: `reqwest::Error` just wraps `serde_json::Error` in this case

eppo_core/src/event_ingestion/event_delivery.rs
rasendubi: The `?` in `let full = self.bytes().await?;` may return an error. The implementation is somewhat convoluted to explain quickly but it may eventually jump to [here](https://docs.rs/reqwest/latest/src/reqwest/async_impl/body.rs.html#277) and return an error of the ["body" kind](https://docs.rs/reqwest/latest/reqwest/struct.Error.html#method.is_body), which is essentially "failed to receive body."

rasendubi:
> We thought about it, but didn't think it was appropriate to perform JSON serialization up front when the user calls track and potentially slow it down, since we want that call to be fast. It's a tradeoff for sure. We're making an assumption that this won't trigger for 99%+ for events, so it doesn't make sense to pay that cost in 100% of them, if that makes sense.

Once we get to integrating event ingestion into Python/Ruby SDKs, we'll hit the task of bringing Python/Ruby values into Rust world. At which point, we'll need to copy all the data and will likely resort to [serde-pyobject](https://crates.io/crates/serde-pyobject) or [serde-magnus](https://github.com/OneSignal/serde-magnus). So it's very tempting to serialize events to json immediately.

PR #178, opened 2025-01-24
chore(events): Send only retriable events to `retry` component

felipecsl:
## Description

Addressing feedback here https://github.com/Eppo-exp/eppo-multiplatform/pull/158#discussion_r1926872837 from @rasendubi 

## Testing

Existing tests

### DIFF
--- a/eppo_core/src/event_ingestion/delivery.rs
+++ b/eppo_core/src/event_ingestion/delivery.rs
@@ -54,6 +54,7 @@ impl QueuedBatch {
 
 pub(super) async fn delivery(
     mut uplink: mpsc::Receiver<BatchedMessage<QueuedEvent>>,
+    retry_downlink: mpsc::Sender<BatchedMessage<QueuedEvent>>,
     delivery_status: mpsc::Sender<QueuedBatch>,
     max_retries: u32,
     event_delivery: EventDelivery,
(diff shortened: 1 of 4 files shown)


eppo_core/src/event_ingestion/delivery.rs
```diff
@@ -130,9 +138,3 @@ fn collect_delivery_response(
         retry,
     }
 }
-
-async fn deliver_status(receiver: &mpsc::Sender<QueuedBatch>, status: QueuedBatch) {
-    receiver.send(status).await.unwrap_or_else(|err| {
-        warn!("Failed to send delivery status: {}", err);
-    });
-}
```
felipecsl, line 138: I thought this was unnecessary so just cleaned up

felipecsl:
> heh, that didn't became much simpler as I expected because `delivery` still cares about `attempts`. But this PR is a step forward and unlocks further refactorings
> 
> I just now took the time to understand what `retry` actually does, and it seems to send failed events almost full circle. There's a subtle bug in bypassing auto-flusher (this could cause the batch to never get retried if user doesn't insert new events, in which case the batcher may hold off the batch indefinitely because auto-flusher won't know to flush).

@rasendubi we don't send retries back to the batcher, it's sent to `batcher_downlink_tx`, which means they land back in the `delivery` component, if I understand correctly

rasendubi:
oh, you're right! I think that's better then (we need to find a better way for naming those channels though)

We're probably susceptible to a deadlock then:
- `delivery` has full uplink from `batcher`
- `delivery` sends 2 batches to `retry`. Sending second batch blocks because the channel is full
- `retry` processes a batch and tries to send it back to `delivery`. It gets blocked because `delivery` uplink is full

`delivery` and `retry` are now both blocked and are unable to make progress.

We can solve this easily by using unbounded channel from `delivery` to `retry`.

rasendubi:
Actually, about naming... what do you think about stripping tx/rx and just using "uplink"/"downlink" (they actually already encode direction — uplink=rx, downlink=tx)

```rust
let (input, flusher_uplink) = mpsc::channel(config.max_queue_size);
let (flusher_downlink, batcher_uplink) = mpsc::channel(1);
let (batcher_downlink, delivery_uplink) = mpsc::channel(1);
let (delivery_downlink, retry_uplink) = mpsc::channel(1);
let (retry_downlink, output) = mpsc::channel(1);
```

felipecsl:
> Actually, about naming... what do you think about stripping tx/rx and just using "uplink"/"downlink" (they already encode direction — uplink=rx, downlink=tx)
> 
> ```rust
> let (input, flusher_uplink) = mpsc::channel(config.max_queue_size);
> let (flusher_downlink, batcher_uplink) = mpsc::channel(1);
> let (batcher_downlink, delivery_uplink) = mpsc::channel(1);
> let (delivery_downlink, retry_uplink) = mpsc::channel(1);
> let (retry_downlink, output) = mpsc::channel(1);
> ```

I never liked the `tx`/`rx` notation, so I'm all up for it. As per the deadlock, that sounds about right. Wanna send a PR with both the renames and the unbounded channel?

PR #181, opened 2025-01-24
feat(core): add SdkKey type

rasendubi:


felipecsl (commented):
Can you add the same test cases we have in #182 to guarantee compatibility with JS? I *think* we need `URL_SAFE_NO_PAD` instead of `BASE64_STANDARD`

eppo_core/src/sdk_key.rs
```diff
@@ -0,0 +1,96 @@
...
+
+use crate::Str;
+
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct SdkKey(Str);
+
+/// Blank implementation of `Debug`, so we don't accidentally leak SDK key in logs.
+impl std::fmt::Debug for SdkKey {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", "SdkKey{..}")
```
felipecsl, line 11: good thinking

rasendubi:
Good catch on url-safe base64! I assumed js-base64 does uses standard base64 (not url-safe) by default, so I used the same. But it actually accepts both alphabets in decoding and I believe we want the url-safe one. I have updated the PR!

felipecsl (approved):

ploomiz (approved):

PR #182, opened 2025-01-24
feat(ruby): Add stub for public event ingestion API to ruby-sdk

felipecsl:
## Description

Added `track()` which is still a no-op, but wanted to get early feedback on the approach here. 
I wasn't sure how to add `event_dispatcher` to `EppoClient::Core::Client` conditionally only if the `event_ingestion` feature is enabled. Is that possible?

* Add SDK Key ingestion URL parsing and tests (ported from JS implementation at https://github.com/Eppo-exp/js-sdk-common/blob/main/src/events/sdk-key-decoder.ts
* Made `EventDispatcherConfig` and `EventDispatcher` public, so they can be accessed from `ruby-sdk`

## Testing

Wrote tests for decoder but other stuff still untested and unused for the time being.


eppo_core/src/event_ingestion/event_dispatcher.rs
```diff
@@ -13,7 +13,7 @@ const MIN_BATCH_SIZE: usize = 1;
 const MAX_BATCH_SIZE: usize = 10_000;
 
 #[derive(Debug, Clone)]
-pub(super) struct EventDispatcherConfig {
+pub struct EventDispatcherConfig {
```
felipecsl, line 16: i had to make this `pub struct` so it's accessible from `ruby-sdk`

eppo_core/src/event_ingestion/event_dispatcher.rs
```diff
@@ -24,9 +24,31 @@ pub(super) struct EventDispatcherConfig {
...
+        }
+    }
+}
+
+const DEFAULT_BATCH_SIZE: usize = 1_000;
+const DEFAULT_DELIVERY_INTERVAL: Duration = Duration::from_millis(10_000);
+const DEFAULT_RETRY_INTERVAL: Duration = Duration::from_millis(5_000);
+const DEFAULT_MAX_RETRY_DELAY: Duration = Duration::from_millis(30_000);
+const DEFAULT_MAX_RETRIES: u32 = 3;
+const DEFAULT_MAX_QUEUE_SIZE: usize = 10_000;
```
felipecsl, line 47: source: https://github.com/Eppo-exp/js-client-sdk/blob/main/src/index.ts#L855-L862

eppo_core/src/event_ingestion/sdk_key_decoder.rs
```diff
@@ -0,0 +1,61 @@
+use base64::engine::general_purpose::URL_SAFE_NO_PAD;
+use base64::prelude::*;
+
+const PATH: &str = "v0/i";
+
+pub fn decode_event_ingestion_url(sdk_key: &str) -> Option<String> {
```
rasendubi, line 6: Ooops, I think we clashed. I started working on spawning event ingestion on the background runtime and realized I need some public API that would trigger it, so I started working on initialization and config. I quickly discovered that this decoding part is missing so here it is: #181

eppo_core/src/event_ingestion/event_dispatcher.rs
rasendubi, line 16: I would like to make fields private or mark the config as `#[non_exhaustive]`, so we can add more fields without bumping major version

eppo_core/Cargo.toml
```diff
@@ -43,6 +43,7 @@ tokio-util = { version = "0.7.13", default-features = false, features = ["rt"] }
 url = "2.5.0"
 uuid = { version = "1.11.0", features = ["v4", "serde"], optional = true }
 exponential-backoff = { version = "2.0.0", optional = true }
+form_urlencoded = { version = "1.2.1", optional = true }
```
rasendubi, line 46: minor: `url` re-exposes `form_urlencoded`, so we don't need a new dependency

eppo_core/src/event_ingestion/mod.rs
```diff
@@ -4,8 +4,9 @@ mod batcher;
 mod delivery;
 mod event;
 mod event_delivery;
-mod event_dispatcher;
 mod queued_event;
 mod retry;
+pub mod sdk_key_decoder;
+pub mod event_dispatcher;
```
rasendubi, line 10: I prefer having "flat" API — instead of making the whole submodule public, re-export public items with `pub use`.

I found that this has a couple benefits:
- tighter control of our API (so less unexpected breaking changes)
- nicer import path for users

eppo_core/src/event_ingestion/sdk_key_decoder.rs
rasendubi, line 6: `decode_event_ingestion_url` should probably be private to core

ruby-sdk/lib/eppo_client/client.rb
```diff
@@ -43,6 +43,10 @@ def shutdown
       @core.shutdown
     end
 
+    def track(event_type, payload)
+      @core.track(event_type, payload)
+    end
```
rasendubi, line 48: Is there a way to mark ruby functions as unstable? I would like to do that, so we can keep treating API as unstable and iterate on it without breaking changes

eppo_core/src/event_ingestion/event_dispatcher.rs
```diff
@@ -24,9 +24,31 @@ pub(super) struct EventDispatcherConfig {
...
+
 /// EventDispatcher is responsible for batching events and delivering them to the ingestion service
 /// via [`EventDelivery`].
-pub(super) struct EventDispatcher {
+pub struct EventDispatcher {
```
rasendubi, line 51: I would like to fix `EventDispatcher` type before we make it public.

It's currently meaningless because it's equal to `EventDispatcherConfig`. I would like its constructor to spawn it (similar to `spawn_configuration_poller`) and its impl to expose a nicer API (`track()` instead of a pair of channels). (These two should make integration into SDKs much simpler.)

eppo_core/src/event_ingestion/sdk_key_decoder.rs
felipecsl: Cool no worries, sounds like we started working on the same stuff :)

felipecsl:
@rasendubi this will need a review from you, even though you wrote a good chunk of this code :)

ruby-sdk/lib/eppo_client/client.rb
rasendubi: @felipecsl I have resolved all of my comments except this one. What do you think about prefixing this function with `unstable_`?
```suggestion
    def unstable_track(event_type, payload)
      @core.track(event_type, payload)
    end
```

eppo_core/src/lib.rs
```diff
@@ -80,4 +80,5 @@ pub use attributes::{
 };
 pub use configuration::Configuration;
 pub use error::{Error, EvaluationError, Result};
+pub use sdk_key::SdkKey;
```
rasendubi, line 83: This technically needs a minor version bump unless we feature-gate it under `event_ingestion`

ruby-sdk/lib/eppo_client/client.rb
felipecsl: hmmm I'm hesitant about it. i'm not aware of any built-in ruby features for marking as unstable. I'd say a comment about it should suffice, also as long as it's not yet in the public docs, then it in general shouldn't be used, but I think I'll defer to @leoromanovsky here

rasendubi:
I want to keep `main` in releasable state, so I'm a bit cautious approving this PR. @felipecsl do you think merging this PR would block a release, or you would be OK shipping it to prod?

eppo_core/src/lib.rs
felipecsl: we probably should gate it with others since this should only be used in the `event_ingestion` codepath

ruby-sdk/ext/eppo_client/src/client.rs
```diff
@@ -237,8 +250,26 @@ impl Client {
...
         }
     }
+
+    pub fn track(&self, event_type: String, payload: Value) -> Result<()> {
+        let Some(event_ingestion) = &self.event_ingestion else {
+            // Event ingestion is disabled, do nothing.
+            return Ok(());
+        };
+
+        let payload: serde_json::Value = serde_magnus::deserialize(payload).map_err(|err| {
```
felipecsl, line 264: since we're parsing the `payload` as json here, it's easy to move the event validation logic from `EventDelivery` to here now. I'll tackle that next

ruby-sdk/lib/eppo_client/client.rb
felipecsl: renamed

rasendubi (approved):

PR #183, opened 2025-01-27
feat(core): tie up background runtime, event ingestion, and tracking

rasendubi:
Based on top of both #182 and #181. (This PR is targeting #182.)

This implements spawning of event ingestion tasks on the background thread, and ties up Ruby SDKs track with event ingestion. Event ingestion from Ruby should work now (although I haven't tested it).


eppo_core/src/event_ingestion/event_dispatcher.rs
rasendubi, line 1: This file has been gradually refactored to `event_ingestion.rs` but there's quite a few changes, so git doesn't pick it up as a rename.

eppo_core/src/event_ingestion/event_ingestion.rs
rasendubi, line 1: This file has been gradually refactored from `event_dispatcher.rs` but there's quite a few changes, so git doesn't pick it up as a rename.

eppo_core/src/event_ingestion/event_ingestion.rs
```diff
@@ -0,0 +1,228 @@
...
+        Mock::given(method("POST"))
+            .and(path("/"))
+            .and(body_json(&json!({
+                "eppo_events": [event.clone()],
+            })))
+            .and(header("x-eppo-token", "test-sdk-key"))
+            .respond_with(ResponseTemplate::new(200).set_body_json(json!({
+                "failed_events": []
+            })))
+            .expect(1)
```
rasendubi, line 160: The tests have been refactored to rely on wiremock's verification (simpler tests + nicer failure messages).

eppo_core/src/event_ingestion/retry.rs
```diff
@@ -1,40 +1,34 @@
...
+            {
                 continue;
             }
+
+            retry_downlink.send(msg).await.ok()?;
         }
-        delivery_status
-            .send(QueuedBatch::retry(batch))
-            .await
-            .ok()?;
```
rasendubi, line 37: Realized that it doesn't make sense for `retry` to send delivery statuses because `delivery` already sent those retries.

There was also a small bug here: `retry` was exiting if nobody was listening for delivery statuses (it should have ignored errors here)

This also allows us to avoid copying the batch (on line 27)

eppo_core/src/event_ingestion/event_ingestion.rs
```diff
@@ -0,0 +1,228 @@
...
+    /// Starts the event ingestion subsystem on the given background runtime.
+    pub fn spawn(runtime: &BackgroundRuntime, config: &EventIngestionConfig) -> EventIngestion {
+        let event_delivery =
+            EventDelivery::new(config.sdk_key.clone(), config.ingestion_url.clone());
+
+        let (input, flusher_uplink) = mpsc::channel(config.max_queue_size);
+        let (flusher_downlink, batcher_uplink) = mpsc::channel(1);
+        let (batcher_downlink, delivery_uplink) = mpsc::channel(1);
+        let (delivery_downlink, retry_uplink) = mpsc::channel(1);
+        let (delivery_status_tx, delivery_status_rx) = mpsc::channel(1);
```
felipecsl, line 73: @rasendubi in a different PR you mentioned you wanted to bump one of these to avoid deadlocks, still planning it for the future?

eppo_core/src/event_ingestion/event_ingestion.rs
```diff
@@ -0,0 +1,228 @@
...
+            retry_uplink,
+            // This is sending to input of delivery
+            batcher_downlink,
+            config.max_retries,
+            config.retry_interval,
+            config.max_retry_delay,
+        ));
+
+        // For now, nobody is interested in delivery statuses.
+        let _ = delivery_status_rx;
```
felipecsl, line 102: i was using this in tests to assert on the output of the "pipeline" but if you think the wiremock assertions are sufficient, that works for me

eppo_core/src/event_ingestion/event_ingestion.rs
```diff
@@ -0,0 +1,228 @@
...
+
+        let mock_server = MockServer::start().await;
+        Mock::given(method("POST"))
+            .and(path("/"))
+            .and(body_json(&json!({"eppo_events": [event] })))
+            .and(header("x-eppo-token", "test-sdk-key"))
+            .respond_with(ResponseTemplate::new(200).set_body_json(json!({
+                "failed_events": [event.uuid],
+            })))
+            .expect(3) // 1 regular attempt + 2 retries
```
felipecsl, line 183: nice but i kinda miss introspecting into the internal state of the channels. Consider exposing the received like before to keep that, unless you have a strong reason not to

eppo_core/src/event_ingestion/retry.rs
felipecsl: nice find!

ruby-sdk/ext/eppo_client/src/client.rs
```diff
@@ -245,19 +250,26 @@ impl Client {
...
     pub fn shutdown(&self) {
-        if let Some((thread, _poller)) = self.poller_thread.take() {
-            let _ = thread.graceful_shutdown();
+        if let Some(thread) = self.background_thread.take() {
+            thread.graceful_shutdown();
         }
     }
 
     pub fn track(&self, event_type: String, payload: Value) -> Result<()> {
+        let Some(event_ingestion) = &self.event_ingestion else {
```
felipecsl, line 259: 🚀

ruby-sdk/ext/eppo_client/src/client.rs
```diff
@@ -107,22 +113,21 @@ impl Client {
...
             None
         };
 
-        let evaluator = Evaluator::new(EvaluatorConfig {
-            configuration_store: configuration_store.clone(),
-            sdk_metadata: SDK_METADATA,
-        });
+        let event_ingestion = config
+            .event_ingestion_config
+            .map(|config| config.spawn(background_thread.runtime()));
```
felipecsl, line 123: in what scenario will `event_ingestion` be `None`? if SDKKey fails to parse the ingestion config?

felipecsl (approved):
looks awesome!

eppo_core/src/event_ingestion/event_ingestion.rs
rasendubi: Yes! It needs to be unbound, which is a different channel type, so needs changes to both retry and delivery components.

I also have a thought for another solution. Given how simple retry technically is, we could just spawn one-off delay-then-send tasks on tokio runtime directly from delivery, and that would remove the loop (and thus deadlock) completely

eppo_core/src/event_ingestion/event_ingestion.rs
rasendubi: The main issue is that nobody is reading that channel in production code yet, so it clogs and blocks the sending part, so everything halts.

Also, because it's a single-consumer channel, we won't be able to read it in tests as soon as we introduce persistence.

I agree that it's useful for testing but the tests that rely on it _are_ peeking into implementation details, so shall be lower-level tests (e.g., arrange a couple of components and test that they work as expected).

ruby-sdk/ext/eppo_client/src/client.rs
rasendubi: Correct
https://github.com/Eppo-exp/eppo-multiplatform/blob/c2c54f25f4754fa4b83b5896dcce945febd57dd2/eppo_core/src/event_ingestion/event_ingestion.rs#L35-L38

PR #186, opened 2025-01-28
fix(core): resolve delivery-retry deadlock

rasendubi:
Now instead of having two component form a loop, we handle retries internally in `delivery` component. It creates a new private unbounded channel for retries and selects from both channels when looking for batch to send. Retries are handled inline by spawning a delay+insert task. (The code is probably easier to follow than this explanation 😅)

Ended up with a larger refactoring that fixes a couple of other todos:
- Removed size validation from `EventDelivery` as we're planning to add it earlier in the pipeline
- Removed `QueuedEvent` and all event copies.
- Cleaned up error handling in event delivery, un-clumped json serialization error, added the main non-retriable error codes


eppo_core/src/event_ingestion/delivery.rs
```diff
@@ -50,108 +46,88 @@ impl DeliveryStatus {
...
+            msg = retries_rx.recv() => msg?,
+        };
+
+        let BatchedMessage { batch, flush } = msg;
+
+        let mut result = event_delivery.deliver(batch).await;
+
+        if attempts >= config.max_retries {
+            // Exceeded max retries -> promote retriable errors to permanent ones.
+            result.failure.append(&mut result.retry);
```
rasendubi, line 79: `append()` removes all events from `result.retry` and inserts them into `result.failure`.

eppo_core/src/event_ingestion/delivery.rs
```diff
@@ -50,108 +46,88 @@ impl DeliveryStatus {
...
-                        // non-retriable.
-                        delivery_status
-                            .send(DeliveryStatus::failure(batch))
-                            .await
-                            .ok()?;
-                    }
-                }
-            }
+
+        let retry_batch = std::mem::take(&mut result.retry);
```
rasendubi, line 82: `std::mem::take` replaces `result.retry` with an empty vector and returns all events to us (indiana jones style).

This way, we avoid cloning the whole batch of events for a retry but as a side effect, this means that retried events are not included in the delivery status (which nobody reads anyway for now)

eppo_core/src/event_ingestion/delivery.rs
```diff
@@ -50,108 +46,88 @@ impl DeliveryStatus {
...
-    DeliveryStatus {
-        success,
-        failure,
-        retry,
-    }
+    min_retry_delay: Duration,
+    max_retry_delay: Duration,
+) {
+    // TODO: exponential_backoff iterator interface doesn't really suit our usage. Rewrite backoff
+    // calculation.
```
rasendubi, line 120: Made `wait_exponential_backoff` infallible: if `Backoff` returns `None`, we use `max_retry_delay`. But overall, `exponential_backoff` iterator API is not really suitable for our use case, so I would rather rewrite backoff calculation or look for another library

eppo_core/src/event_ingestion/event_delivery.rs
```diff
@@ -19,202 +19,191 @@ pub(super) struct EventDelivery {
...
-
-#[derive(serde::Deserialize, Debug)]
-pub(super) struct RawEventDeliveryResponse {
-    pub failed_events: HashSet<Uuid>,
-}
-
 #[derive(thiserror::Error, Debug)]
 pub(super) enum EventDeliveryError {
-    #[error("Single event payload too large {0} (expected < {max})", max = MAX_EVENT_SERIALIZED_LENGTH)]
-    EventPayloadTooLargeError(usize),
```
rasendubi, line 50: This is no longer event delivery's concern, so removed

eppo_core/src/event_ingestion/event_delivery.rs
```diff
@@ -19,202 +19,191 @@ pub(super) struct EventDelivery {
...
-    #[error("Failed to serialize events to JSON")]
-    JsonSerializationError(serde_json::Error),
-    #[error("Failed to parse JSON response")]
-    JsonDeserializationError(reqwest::Error),
```
rasendubi, line 54: Clumped these two into non-retriable error

eppo_core/src/event_ingestion/event_delivery.rs
```diff
@@ -19,202 +19,191 @@ pub(super) struct EventDelivery {
...
+                Some(StatusCode::UNAUTHORIZED | StatusCode::FORBIDDEN) => {
+                    log::warn!(target: "eppo", "client is not authorized. Check your API key");
+                    EventDeliveryError::NonRetriableError(err)
+                }
+                Some(
+                    status @ (StatusCode::BAD_REQUEST
+                    | StatusCode::NOT_FOUND
+                    | StatusCode::METHOD_NOT_ALLOWED
+                    | StatusCode::CONFLICT
+                    | StatusCode::UNPROCESSABLE_ENTITY),
```
rasendubi, line 46: Pulled this list of non-retriable errors from a random list on the internet

eppo_core/src/event_ingestion/event_ingestion.rs
```diff
@@ -7,30 +7,32 @@ use uuid::Uuid;
...
     pub ingestion_url: Url,
+    pub max_queue_size: usize,
     pub delivery_interval: Duration,
-    pub retry_interval: Duration,
-    pub max_retry_delay: Duration,
-    pub max_retries: u32,
     pub batch_size: usize,
-    pub max_queue_size: usize,
+    pub max_retries: u32,
+    pub base_retry_delay: Duration,
```
rasendubi, line 25: The naming of `retry_interval` and `max_retry_delay` didn't quite connect. I renamed `retry_interval` to `base_retry_delay`, so I hope `base_retry_delay` + `max_retry_delay` is easier to understand

eppo_core/src/event_ingestion/event_ingestion.rs
```diff
@@ -7,30 +7,32 @@ use uuid::Uuid;
...
+    pub max_retry_delay: Duration,
```
rasendubi, line 26: Reordered fields mostly in the order where they apply in the pipeline

rasendubi:
Love it when PRs are net negative:
image

eppo_core/src/event_ingestion/delivery.rs
```diff
@@ -50,108 +46,88 @@ impl DeliveryStatus {
...
-            batch,
-            flush: _flush,
-        } = uplink.recv().await?;
-        if batch.is_empty() {
-            continue;
+        // Randomly select between sending a new batch or retrying an old one.
+        let (attempts, msg) = tokio::select! {
+            msg = uplink.recv() => (0, msg?),
+            msg = retries_rx.recv() => msg?,
+        };
```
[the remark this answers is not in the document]
felipecsl: also can you expand on the "randomly" portion? if both channels already have a pending message for receiving by the time we call `select!`, then it will pick one of the two at random?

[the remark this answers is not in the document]
rasendubi: We unpack the result of select into `(attempts, msg)`, so `(0, msg?)` is setting attempts counter for messages coming from uplink.

`?` is to kill delivery service if uplink died (if `msg` is error). I think we'll need to change that behavior once we get to graceful shutdown.

re "randomly": yes, `select` is fair in a sense that if both channels are ready, it will pick one at random

eppo_core/src/event_ingestion/event_delivery.rs
```diff
@@ -19,202 +19,190 @@ pub(super) struct EventDelivery {
...
+                    EventDeliveryError::NonRetriableError(err)
+                }
+                Some(status) if status.is_server_error() => {
+                    log::warn!(target: "eppo", "received {status} response delivering events: {err:?}");
+                    EventDeliveryError::RetriableError(err)
+                }
+                _ => {
+                    // Other error statuses **might be** retriable
+                    log::warn!(target: "eppo", "received non-200 response delivering events: {:?}", err);
+                    EventDeliveryError::RetriableError(err)
```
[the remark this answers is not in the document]
rasendubi: What do you mean by final error? I think I pulled the list of most important errors that shouldn't be retried. It should be fine to retry the rest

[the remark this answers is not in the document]
felipecsl: by "final" i mean non-retriable. I think this should cover our basis, so we're fine, minor nit/non blocking

PR #195, opened 2025-02-04
chore(ci): rely on status code instead of logs

rasendubi:


.github/workflows/publish.yml
```diff
@@ -196,8 +196,8 @@ jobs:
         if: matrix._.platform == 'x86_64-linux' # GitHub actions architecture
         run: |
           gem install pkg/eppo-server-sdk-*.gem --verbose
-          script="puts EppoClient::Core::Client.new(EppoClient::Config.new('placeholder'))"
-          ruby -reppo_client -e "$script" 2>&1 | grep 'fetching new configuration'
+          script="EppoClient::init(EppoClient::Config.new('placeholder')); EppoClient::Client.instance.shutdown"
```
rasendubi, line 199: `EppoClient::Core` is an internal module.

rasendubi, line 199: Decided to wait for `shutdown` as it's calling through to Rust

felipecsl (approved):
thank you

.github/workflows/publish.yml
```diff
@@ -196,8 +196,8 @@ jobs:
...
+          ruby -reppo_client -e "$script" 2>&1 || { echo "❌ Failed to run smoke test"; exit 1 }
```
rasendubi, line 200: We have removed that log line and CI started failing with a false negative. It would make more sense if that was checked in PRs, so it could be fixed quickly but right now the check is skipped in PRs and fails when we decide to release, which is annoying (as we have to fix the CI when we don't expect and jump through the hoops of yanking tags).

I want something more stable for a pre-release check, so I propose to check the exit code instead (but exercise a bit more code)

rasendubi, line 200: Also added an echo as currently the CI logs don't explain the failure reason and are kind of cryptic:
image

PR #197, opened 2025-02-05
FF-3753 feat(dart): Dart SDK

rasendubi:


rasendubi:
I'm going to work on adding documentation next but this is now ready for review 🎉

eppo_core/src/background/runtime.rs
```diff
@@ -101,6 +101,7 @@ impl BackgroundRuntime {
 
     /// Command background activities to stop and exit.
     pub(crate) fn stop(&self) {
+        log::debug!(target: "eppo", "stopping background runtime");
```
schmit, line 104: Did you intend to keep this debug here?

rasendubi: Absolutely: https://github.com/Eppo-exp/eppo-multiplatform/pull/197/commits/5353290c3de41dfbd6a6d61f66800673c3cb6e4b

I've spent a couple of days debugging why poller mysteriously crashes. That log line would cut the debugging to one minute 🙈

dart-sdk/lib/src/client.dart
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  /// Returns a Future that resolves when the client completes initialization,
+  /// receives a configuration, and is ready to serve assignment and bandits.
+  Future<void> whenReady() async {
+    await this._initialized;
+    await this._core!.waitForInitialization();
+  }
+
+  String stringAssignment(String flagKey, Subject subject, String defaultValue) {
```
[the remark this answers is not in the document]
rasendubi: Yeah, the ask for review leaked slightly earlier than I expected, so I didn't have the time to annotate all the interesting parts 🙂

For the difference here: I hate `Attributes`/`ContextAttributes` distinction in other SDKs, so I tried to experiment with it. This change makes signature uniform for both assignments and bandits. Bandit signature for comparison (no `ContextAttributes`!):
```dart
EvaluationResult<String> banditAction(
  String flagKey,
  Subject subject,
  Map<String, Attributes> actions,
  String defaultVariation)
```

and example usage in Dart (from test):
```dart
      final bandit = client.banditAction(
        'banner_bandit_flag',
        Subject('alice')
          ..numberAttribute('age', 25)
          ..stringAttribute('country', 'USA')
          ..stringAttribute('genderIdentity', 'female'),
        {
          'nike': Attributes()
            ..numberAttribute('brand_affinity', 1.0)
            ..stringAttribute('loyalty_tier', 'gold')
            ..boolAttribute('purchased_last_30_days', true),
        },
        'control',
      );
```

That seems to work better for typed languages that don't allow mixing strings and numbers in the same container. I like it but curious what other think.

package.json
```diff
@@ -8,6 +8,7 @@
     "test:rust": "cargo test",
     "test:python": "cd python-sdk && pytest",
     "test:ruby": "cd ruby-sdk && bundle exec rake build && bundle exec rspec",
+    "test:dart": "cd dart-sdk && dart --enable-experiment=native-assets test",
```
leoromanovsky, line 11: Tried running this locally with exception:

```
   Compiling eppo_dart v1.0.0 (/Users/leo/src/eppo-multiplatform/dart-sdk/rust)
error: failed to run custom build command for `eppo_dart v1.0.0 (/Users/leo/src/eppo-multiplatform/dart-sdk/rust)`

Caused by:
  process didn't exit successfully: `/Users/leo/src/eppo-multiplatform/dart-sdk/rust/target/release/build/eppo_dart-6f7c55cf4f81c058/build-script-build` (exit status: 101)
  --- stdout
  [2025-02-22T16:06:52.693Z DEBUG /Users/leo/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.8.0/src/library/codegen/config/config_parser.rs:51] Found config file ../flutter_rust_bridge.yaml
  [2025-02-22T16:06:52.694Z DEBUG /Users/leo/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.8.0/src/library/codegen/mod.rs:23] config=Config { base_dir: Some(".."), rust_input: Some("crate::api"), dart_output: Some("lib/src/rust"), c_output: Some("frb_generated.h"), duplicated_c_output: None, rust_root: None, rust_output: None, dart_entrypoint_class_name: None, dart_format_line_length: None, dart_preamble: None, rust_preamble: Some("use eppo_core::{Str, AttributeValue};\n"), dart_enums_style: None, add_mod_to_lib: None, llvm_path: None, llvm_compiler_opts: None, dart_root: None, build_runner: None, extra_headers: None, web: None, deps_check: None, dart3: None, full_dep: None, local: None, default_external_library_loader_web_prefix: None, dart_type_rename: None, enable_lifetime: None, type_64bit_int: None, default_dart_async: Some(false), stop_on_error: None, dump: None, dump_all: None, rust_features: None } meta_config=MetaConfig { watch: false }
  [2025-02-22T16:06:52.695Z DEBUG /Users/leo/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.8.0/src/library/codegen/config/internal_config_parser/mod.rs:33] InternalConfig.parse base_dir="/Users/leo/src/eppo-multiplatform/dart-sdk"
  [2025-02-22T16:06:52.696Z ERROR /Users/leo/.cargo/registry/src/index.crates.io-6f17d22bba15001f/flutter_rust_bridge_codegen-2.8.0/src/library/utils/logs.rs:54] panicked at build.rs:25:6:
  called `Result::unwrap()` on an `Err` value: Fail to canonicalize path="/Users/leo/src/eppo-multiplatform/dart-sdk/lib/src/rust"

  Caused by:
      No such file or directory (os error 2)

  --- stderr
  thread 'main' panicked at build.rs:25:6:
  called `Result::unwrap()` on an `Err` value: Fail to canonicalize path="/Users/leo/src/eppo-multiplatform/dart-sdk/lib/src/rust"

  Caused by:
      No such file or directory (os error 2)
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Unhandled exception:
ProcessException: Bad exit code (101). If you want to see extra information, set FRB_DART_RUN_COMMAND_STDERR=1
  Command: cargo build --release
#0      runCommand (package:flutter_rust_bridge/src/cli/run_command.dart:67:5)
<asynchronous suspension>
#1      simpleBuild (file:///Users/leo/src/eppo-multiplatform/dart-sdk/simple_build.dart:54:5)
<asynchronous suspension>

stdout:


Error: Compiling native assets failed.
```

do you expect this command to succeed on macos?

eppo_core/src/background/runtime.rs
```diff
@@ -34,73 +35,70 @@ pub struct BackgroundRuntime {
...
+}
+
+impl<AR> BackgroundRuntime<AR> {
+    /// Returns a cancellation token that get cancelled when background runtime is going to
+    /// shutdown.
+    ///
+    /// If you want your task to react to shutdown and perform some cleanup, spawn your task with
+    /// [`BackgroundThread::spawn_tracked()`] and watch this cancellation token.
+    ///
+    /// Note: this is a child token of the main cancellation token, so canceling it does not cancel
```
leoromanovsky, line 92: very nice - this will be primarily used within SDKs or can you imagine a customer facing use too?

dart-sdk/rust/src/runtime.rs
```diff
@@ -0,0 +1,36 @@
...
+
+#[cfg(not(target_arch = "wasm32"))]
+mod inner {
+    pub(crate) type FlutterRustBridgeRuntime = tokio::runtime::Handle;
+
+    pub(crate) fn get_runtime() -> FlutterRustBridgeRuntime {
+        crate::frb_generated::FLUTTER_RUST_BRIDGE_HANDLER
+            .async_runtime()
+            .0
+             .0
```
leoromanovsky, line 11: If the underlying interfaces changes do you think the compiler will break here and alert us? Consider how to write this with protection and a log message to help a future developer.

eppo_core/src/background/async_runtime.rs
```diff
@@ -0,0 +1,20 @@
+use std::future::Future;
+
+/// `AsyncRuntime` abstracts over various Rust's async runtimes with minimal interface. This is
+/// usually just a tokio Handle but Dart SDK in wasm uses a custom implementation.
```
leoromanovsky, line 4: I did most of my reading around this change to understand it and got introduced to the `Send` interface, thanks! https://doc.rust-lang.org/nomicon/send-and-sync.html

Also needed help from AI to understand how it fit together with the wasm single threaded runtime. 

```
In the future, if more non-Send futures need handling, you may need a more generalized solution (e.g., perhaps relax the trait bound for wasm via conditional compilation, or provide a separate method for non-Send tasks). For now, the approach is correct and contained to config polling.
```

What do you imagine doing down the line to evolve this interface?

dart-sdk/rust/src/api/client.rs
```diff
@@ -0,0 +1,217 @@
...
+            &background,
+            ConfigurationFetcher::new(ConfigurationFetcherConfig {
+                base_url,
+                api_key: sdk_key,
+                sdk_metadata: SDK_METADATA,
+            }),
+            configuration_store.clone(),
+            ConfigurationPollerConfig {
+                interval: Duration::from_secs(30),
+                jitter: Duration::from_secs(3),
```
leoromanovsky, line 57: params needed here?

dart-sdk/rust/src/api/client.rs
```diff
@@ -0,0 +1,217 @@
...
+        let _result = self.poller.wait_for_configuration().await;
+    }
+
+    #[frb(sync, positional)]
+    pub fn string_assignment(
+        &self,
+        flag_key: &str,
+        subject_key: Str,
+        subject_attributes: HashMap<Str, AttributeValue>,
+    ) -> (Option<Str>, Option<AssignmentEvent>) {
```
leoromanovsky, line 84: Is it easier to return this type instead of a `Result` for use in dart or did something else inform your decision?

leoromanovsky (commented):
@rasendubi this is huge - I did a line by line review and enjoying reading all of it. I can tell you had to solve several challenges related to the background runtime across wasm/non-wasm as well as running in the javascript environment. Very appreciated that you didn't take shortcuts but seemed to step back and create generalized interfaces for the new behavior.

Before approving and releasing could you please help me run the dart & flutter web artifacts locally so I can get comfort with the tooling? Apologize for not figuring this out myself after some research.

package.json
rasendubi: oh, it might be because you don't have `lib/src/rust` directory. It's auto-generated, so git-ignored. I've just added a .gitkeep file, so git will create a directory for you when you pull latest changes

eppo_core/src/background/runtime.rs
rasendubi: I don't think there's customer use. Right now, this is only used within core (not even exported to individual sdks)

dart-sdk/rust/src/runtime.rs
rasendubi: Yes, absolutely. That's one of the things that broke in web build because implementation is different.

We are pinning `rust_flutter_bridge` version, so this should not break unexpectedly. And when it does, the error will be scoped to this file, so it will be quite clear what happened

dart-sdk/rust/src/api/client.rs
rasendubi: Good catch! We have the params — just need to use their values

eppo_core/src/background/async_runtime.rs
rasendubi: Yeah, I don't really like this `AsyncRuntime` and I kinda used an escape hatch in configuration poller by invoking wasm spawn directly.

Having different definition for wasm is possible but it's going to be contagious as interfaces of `spawn_tracked`/`_untracked` would change as well. I don't think it's going to propagate much further though, so probably worth biting that bullet at some point.

I'm also not entirely sure what's multi-threading story for wasm is. It looks like there is some support for threading on some platforms and on web via [`SharedArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer) (when cross-origin isolation is enabled).

dart-sdk/rust/src/api/client.rs
rasendubi: Don't remember exactly but it's probably because we needed to convert from uni-typed `AssignmentValue` to `Str`, so ended up with this destructuring:
```rust
        let Ok(Some(Assignment {
            value: AssignmentValue::String(result),
            event,
        })) = self.evaluator.get_assignment(...)
        else {
            return ...;
        };
```

at which point the two sensible options are either `Option<(Str, Option<AssignmentEvent>)>` (or `Result` outside) or `(Option<Str>, Option<AssignmentEvent>)`. The first one is more faithful (because event can only be present if there's an assignment) but the second is slightly easier to work from Dart:
```dart
      final (result, event) = core.stringAssignment(flagKey, subject.key, subject.attributes.coreAttributes);
      this.logAssignmentEvent(event);
      return result ?? defaultValue;
```
vs. having to unnest two levels of optionals

rasendubi:
@leoromanovsky sharing my stashed commands for web build (my understanding is that we'll put them in CI before release):
```
prepare:
    cargo install -f flutter_rust_bridge_codegen@=2.8.0 wasm-pack wasm-bindgen-cli

build-web:
    flutter_rust_bridge_codegen build-web --wasm-pack-rustflags '--cfg getrandom_backend="wasm_js" -C target-feature=+atomics,+bulk-memory,+mutable-globals'
```

rasendubi:
Another piece that I totally forgot to mention is that dart-sdk has its own cargo workspace, so for local testing we need to override eppo_core crate (similarly to how we do in ruby).

Put the following in `dart-sdk/rust/.cargo/config.toml`:
```toml
[patch.crates-io]
eppo_core = { path = '../../eppo_core' }
```

leoromanovsky (approved):
Was a pleasure digging into this! The new async interface is very slick and major kudos for figuring out how to bridge from flutter to rust.

I tested locally by:

1️⃣  flutter web package and running the dart test.

```
flutter_rust_bridge_codegen build-web --wasm-pack-rustflags '--cfg getrandom_backend="wasm_js" -C target-feature=+atomics,+bulk-memory,+mutable-globals'

    Finished `dev` profile [unoptimized + debuginfo] target(s) in 7.90s
[INFO]: Optional fields missing from Cargo.toml: 'description', 'repository', and 'license'. These are not necessary, but recommended
[INFO]: ✨   Done in 9.17s
[INFO]: 📦   Your wasm pkg is ready to publish at /Users/leo/src/eppo-multiplatform/dart-sdk/web/pkg.
```

2️⃣ running the dart test

```
> yarn run start-mock-server
> yarn test:dart
```

Screenshot 2025-02-24 at 9 10 56 PM

dart-sdk/lib/src/subject.dart
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  void numberAttribute(String key, double value, {AttributeKind kind = AttributeKind.numeric}) {
+    this.attributes.numberAttribute(key, value, kind: kind);
+  }
+
+  void boolAttribute(String key, bool value) {
+    this.attributes.boolAttribute(key, value);
+  }
+}
```
leoromanovsky, line 24: This is interesting by the way - are you experimenting with a new API to combine the subject key and attributes or is this canonical dart?

rasendubi: I figured that Dart has named arguments, so I tried to figure out what a good API would look like. I don't want users naming all arguments as that's a bit too cumbersome.

`stringAssignment` signature is the most confusing one with our regular approach:
```
String stringAssignment(String, String, Map<String, dynamic>, String)
```
(good luck figuring out what these strings refer to)

Reducing it to: `String stringAssignment(String, Subject, String)` makes it easier to understand, I guess.

Another point is that in Dart it's preferred to have static typing for all things and Dart doesn't have an ad-hoc union type (like `string | number | boolean` in typescript), so we have to add our own `Attribute` type (or something similar). And I actually hate the distinction between `Attributes` in assignments and `ContextAttributes` in bandits, so I tried to get rid of it.

I think I implemented `Subject` first but then figured that we also need attributes for actions, so I extracted `Attributes` as a separate type, too. I think unsplitting Subject back to `stringAssignment(String, String, Attributes, String)` also works now.

But yeah, very much experimenty, so would like to gather feedback about how this feels from user's pov

dart-sdk/rust/src/api/client.rs
rasendubi: UPD: fixed

felipecsl:
I couldn't get `dart test` to pass yet. any of you run into this?

```
dart --enable-experiment=native-assets test                                                                                                                                                                                                           "/Users/felipecsl/prj/eppo-multiplatform/dart-sdk/hook/build.dart" file not found.
Building native assets for package:eppo_sdk failed.
Compilation of hook returned with exit code: 255.
To reproduce run:
/Users/felipecsl/prj/flutter/flutter/bin/cache/dart-sdk/bin/dart compile kernel --packages=/Users/felipecsl/prj/eppo-multiplatform/dart-sdk/.dart_tool/package_config.json --output=/Users/felipecsl/prj/eppo-multiplatform/dart-sdk/.dart_tool/native_assets_builder/eppo_sdk/36bad3131645e96b3e5f196e7b126c8d/hook.dill --depfile=/Users/felipecsl/prj/eppo-multiplatform/dart-sdk/.dart_tool/native_assets_builder/eppo_sdk/36bad3131645e96b3e5f196e7b126c8d/hook.dill.d /Users/felipecsl/prj/eppo-multiplatform/dart-sdk/hook/build.dart
stderr:
"/Users/felipecsl/prj/eppo-multiplatform/dart-sdk/hook/build.dart" file not found.

stdout:


Error: Compiling native assets failed.
```

leoromanovsky:
> I couldn't get `dart test` to pass yet. any of you run into this?
> 
> ```
> dart --enable-experiment=native-assets test                                                                                                                                                                                                           "/Users/felipecsl/prj/eppo-multiplatform/dart-sdk/hook/build.dart" file not found.
> Building native assets for package:eppo_sdk failed.
> Compilation of hook returned with exit code: 255.
> To reproduce run:
> /Users/felipecsl/prj/flutter/flutter/bin/cache/dart-sdk/bin/dart compile kernel --packages=/Users/felipecsl/prj/eppo-multiplatform/dart-sdk/.dart_tool/package_config.json --output=/Users/felipecsl/prj/eppo-multiplatform/dart-sdk/.dart_tool/native_assets_builder/eppo_sdk/36bad3131645e96b3e5f196e7b126c8d/hook.dill --depfile=/Users/felipecsl/prj/eppo-multiplatform/dart-sdk/.dart_tool/native_assets_builder/eppo_sdk/36bad3131645e96b3e5f196e7b126c8d/hook.dill.d /Users/felipecsl/prj/eppo-multiplatform/dart-sdk/hook/build.dart
> stderr:
> "/Users/felipecsl/prj/eppo-multiplatform/dart-sdk/hook/build.dart" file not found.
> 
> stdout:
> 
> 
> Error: Compiling native assets failed.
> ```

<img width="2471" alt="Screenshot 2025-02-26 at 12 14 09 PM" src="https://github.com/user-attachments/assets/f8f16658-1107-4cc5-be57-f105586166f5" />

Works for me with latest pull.

felipecsl:
> @felipecsl I just pushed a fix for this: [449bb49](https://github.com/Eppo-exp/eppo-multiplatform/commit/449bb49568603b1f0da2e1d118e83726954bfea8)

now it just goes in an infinite loop, build never finishes
```
warning: use of deprecated method `rand::Rng::gen_range`: Renamed to `random_range`
   --> /Users/felipecsl/prj/eppo-multiplatform/eppo_core/src/configuration_poller.rs:168:53
    |
168 |     Duration::saturating_sub(interval, thread_rng().gen_range(Duration::ZERO..=jitter))
    |                                                     ^^^^^^^^^

warning: use of deprecated method `rand::Rng::gen`: Renamed to `random` to avoid conflict with the new `gen` keyword in Rust 2024.
   --> /Users/felipecsl/prj/eppo-multiplatform/eppo_core/src/precomputed.rs:130:44
    |
130 |             let bytes = rand::thread_rng().gen::<[u8; 16]>();
    |                                            ^^^

warning: method `spawn_tracked` is never used
  --> /Users/felipecsl/prj/eppo-multiplatform/eppo_core/src/background/runtime.rs:59:19
   |
38 | impl<AR: AsyncRuntime> BackgroundRuntime<AR> {
   | -------------------------------------------- method in this implementation
...
59 |     pub(crate) fn spawn_tracked<F>(&self, future: F)
   |                   ^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: associated items `new`, `as_str`, and `event_ingestion_url` are never used
  --> /Users/felipecsl/prj/eppo-multiplatform/eppo_core/src/sdk_key.rs:17:12
   |
15 | impl SdkKey {
   | ----------- associated items in this implementation
16 |     #[inline]
17 |     pub fn new(key: Str) -> SdkKey {
   |            ^^^
...
22 |     pub(crate) fn as_str(&self) -> &str {
   |                   ^^^^^^
...
30 |     pub(crate) fn event_ingestion_url(&self) -> Option<Url> {
   |                   ^^^^^^^^^^^^^^^^^^^

warning: `eppo_core` (lib) generated 7 warnings
warning: unused imports: `Attributes`, `BanditEvent`, and `DEFAULT_BASE_URL`
  --> src/api/client.rs:7:79
   |
7  |     configuration_fetcher::{ConfigurationFetcher, ConfigurationFetcherConfig, DEFAULT_BASE_URL},
   |                                                                               ^^^^^^^^^^^^^^^^
...
13 |     events::{AssignmentEvent, BanditEvent},
   |                               ^^^^^^^^^^^
14 |     ufc::{Assignment, AssignmentValue, VariationType},
15 |     AttributeValue, Attributes, SdkMetadata, Str,
   |                     ^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: unused variable: `s`
  --> src/api/mod.rs:31:32
   |
31 | pub fn decode_assignment_event(s: String) -> AssignmentEvent {
   |                                ^ help: if this is intentional, prefix it with an underscore: `_s`
   |
   = note: `#[warn(unused_variables)]` on by default

warning: unused variable: `s`
  --> src/api/mod.rs:42:28
   |
42 | pub fn decode_bandit_event(s: String) -> BanditEvent {
   |                            ^ help: if this is intentional, prefix it with an underscore: `_s`

warning: fields `configuration_store` and `background` are never read
  --> src/api/client.rs:22:5
   |
21 | pub struct CoreClient {
   |            ---------- fields in this struct
22 |     configuration_store: Arc<ConfigurationStore>,
   |     ^^^^^^^^^^^^^^^^^^^
23 |     background: BackgroundRuntime<FlutterRustBridgeRuntime>,
   |     ^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: `eppo_dart` (lib) generated 4 warnings (run `cargo fix --lib -p eppo_dart` to apply 1 suggestion)
    Finished `release` profile [optimized] target(s) in 0.13s
```

rasendubi:
@felipecsl I just pinned dart to 3.6 in CI and that passed through, so the error seems to be caused by dart 3.7 upgrade indeed

leoromanovsky:
> @felipecsl I just pinned dart to 3.6 in CI and that passed through, so the error seems to be caused by dart 3.7 upgrade indeed

Great find! I am on 3.6 locally and couldn't understand why CI was failing too.

```
✗ dart --version
Dart SDK version: 3.6.0 (stable) (Thu Dec 5 07:46:24 2024 -0800) on "macos_arm64"
```

dart-sdk/pubspec.yaml
```diff
@@ -1,20 +1,20 @@
 name: eppo_sdk
 description: Eppo SDK for Dart
-version: '1.0.0'
+version: '0.1.0-alpha'
```
leoromanovsky, line 3: thank you - a nice touch as we do some iteration to work out the packaging questions

felipecsl:
> I can confirm I'm on dart 3.7 - will try downgrading to 3.6

Just to follow up here, I can confirm I was able to build for web with Dart 3.6

.changeset/giant-eyes-tie.md
```diff
@@ -0,0 +1,7 @@
+---
+"eppo_core": major
+---
+
+Make async runtime abstract.
+
+This introduces an `AsyncRuntime` trait that allows us to abstract over different async runtimes. This is required to support Dart SDK that doesn't use tokio runtime in web build.
```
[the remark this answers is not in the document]
felipecsl: we probably need to add a note here about dropping openssl in favor of rust-tls

dart-sdk/rust/build-android-aarch64.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+export CC_aarch64_linux_android="$NDK_HOME/toolchains/llvm/prebuilt/darwin-x86_64/bin/aarch64-linux-android35-clang"
+export CARGO_TARGET_AARCH64_LINUX_ANDROID_LINKER="$CC_aarch64_linux_android"
+export AR_aarch64_linux_android="$NDK_HOME/toolchains/llvm/prebuilt/darwin-x86_64/bin/llvm-ar"
+export RANLIB="$NDK_HOME/toolchains/llvm/prebuilt/darwin-x86_64/bin/llvm-ranlib"
+
+cargo build --release --target aarch64-linux-android -p eppo_dart
```
felipecsl, line 8: we can probably drop this in favor of https://github.com/cross-rs/cross in the future

.changeset/giant-eyes-tie.md
```diff
@@ -0,0 +1,7 @@
+---
+"eppo_core": major
```
leoromanovsky, line 2: Thanks for making this a major bump - makes a ton of sense to give us room to test python and ruby more extensively.

felipecsl (approved):

.changeset/giant-eyes-tie.md
rasendubi: It doesn't give us much room. It is going to bump eppo_core's version in both python and ruby, so the next python/ruby releases will release with this new change. (This is the same regardless of whether it's a major or patch change.)

.changeset/giant-eyes-tie.md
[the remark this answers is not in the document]
rasendubi: @felipecsl I've added another changeset for this. This one affects all SDKs though

rasendubi:
@felipecsl python build started to fail. Do you think it is related to the rustls changes?

felipecsl:
> @felipecsl python build started to fail. Do you think it is related to the rustls changes?

@rasendubi it was working on friday on my branch #222 and also started failing there, I think something changed in the underlying OS, taking a look

felipecsl:
anything blocking this PR from being merged?

PR #198, opened 2025-02-05
fix(events): Serialize Event `event_type` as `type`

felipecsl:
## Description

This should have been just `type` but used `event_type` instead due to it being a reserved word in Rust, but we need to make sure it's serialized as `type`

## Testing

Wrote a test

FF-3954


.changeset/lemon-baboons-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+"eppo_core": patch
+---
+
+Ingestion: Fix JSON serialization of Event type field
```
rasendubi, line 5: This is going to appear in public release notes / changelog, so want to mark all mentions of event ingestion as "unstable"
```suggestion
[unstable] Event Ingestion: Fix JSON serialization of Event type field
```

.changeset/lemon-baboons-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+"eppo_core": patch
```
rasendubi, line 2: We probably want a Ruby release as well:
```suggestion
"eppo_core": patch
"ruby-sdk": patch
```

eppo_core/src/event_ingestion/event.rs
```diff
@@ -5,6 +5,7 @@ use serde::{Deserialize, Serialize};
 pub(super) struct Event {
     pub uuid: uuid::Uuid,
     pub timestamp: Timestamp,
+    #[serde(rename = "type")]
```
rasendubi, line 8: > eh seems like unnecessary complexity to me — [old felipe](https://github.com/Eppo-exp/eppo-multiplatform/pull/132#discussion_r1902230130) 😅

rasendubi (approved):

eppo_core/src/event_ingestion/event.rs
felipecsl: yup i was wrong :D

PR #201, opened 2025-02-06
fix(events): Ensure Event timestamp is sent as epoch milliseconds

felipecsl:
## Description

Event timestamp field should be sent as epoch milliseconds numeric value instead of datetime string.

## Testing

Wrote a test

FF-3962

### DIFF
--- a/eppo_core/src/event_ingestion/event.rs
+++ b/eppo_core/src/event_ingestion/event.rs
@@ -1,11 +1,39 @@
 use crate::timestamp::Timestamp;
-use serde::{Deserialize, Serialize};
+use chrono::{DateTime, Utc};
+use serde::{Deserialize, Serialize, Serializer};
+use serde::de::Deserializer;
 
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
 pub(super) struct Event {
     pub uuid: uuid::Uuid,
+    #[serde(
+        serialize_with = "serialize_millis",
+        deserialize_with = "deserialize_millis"
+    )]
     pub timestamp: Timestamp,
     #[serde(rename = "type")]
     pub event_type: String,
     pub payload: serde_json::Value,
+}
+
+fn serialize_millis<S>(date: &Timestamp, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    // Convert to i64 representing milliseconds since the Unix epoch
+    let millis = date.timestamp_millis();
+    serializer.serialize_i64(millis)
+}
+
+fn deserialize_millis<'de, D>(deserializer: D) -> Result<Timestamp, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let millis = i64::deserialize(deserializer)?;
+    // Convert i64 millis back to Timestamp
+    let secs = millis / 1000;
+    let subsec_nanos = ((millis % 1000) * 1_000_000) as u32;
+    let naive = DateTime::from_timestamp(secs, subsec_nanos)
+        .ok_or_else(|| serde::de::Error::custom("Invalid timestamp"))?;
+    Ok(Timestamp::from(naive))
 }
--- a/eppo_core/src/event_ingestion/event.rs
+++ b/eppo_core/src/event_ingestion/event.rs
@@ -1,7 +1,7 @@
 use crate::timestamp::Timestamp;
 use chrono::{DateTime, Utc};
-use serde::{Deserialize, Serialize, Serializer};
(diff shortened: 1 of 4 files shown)


eppo_core/src/event_ingestion/event.rs
```diff
@@ -1,11 +1,39 @@
...
+where
+    D: Deserializer<'de>,
+{
+    let millis = i64::deserialize(deserializer)?;
+    // Convert i64 millis back to Timestamp
+    let secs = millis / 1000;
+    let subsec_nanos = ((millis % 1000) * 1_000_000) as u32;
+    let naive = DateTime::from_timestamp(secs, subsec_nanos)
+        .ok_or_else(|| serde::de::Error::custom("Invalid timestamp"))?;
+    Ok(Timestamp::from(naive))
```
rasendubi, line 38: There's a reverse method for `timestamp_millis`: [DateTime::from_timestamp_millis](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.from_timestamp_millis) (I'm too lazy to verify the math here 😅)

`from_timestamp`/`from_timestamp_millis` also return a `DateTime<Utc>` (aka `Timestamp`), so the `Timestamp::from` on the list line is a no-op
```suggestion
    DateTime::from_timestamp_millis(millis)
      .ok_or_else(|| serde::de::Error::custom("Invalid timestamp"))
```

eppo_core/src/lib.rs
```diff
@@ -56,12 +56,12 @@ pub mod configuration_fetcher;
 pub mod configuration_poller;
 pub mod configuration_store;
 pub mod eval;
-mod sdk_key;
 #[cfg(feature = "event_ingestion")]
 pub mod event_ingestion;
 pub mod events;
 #[cfg(feature = "pyo3")]
 pub mod pyo3;
+mod sdk_key;
```
rasendubi, line 64: Could you move it to the block of non-public mods below?

eppo_core/src/event_ingestion/event.rs
```diff
@@ -1,11 +1,39 @@
...
+use serde::{Deserialize, Serialize, Serializer};
 
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
 pub(super) struct Event {
     pub uuid: uuid::Uuid,
+    #[serde(
+        serialize_with = "serialize_millis",
+        deserialize_with = "deserialize_millis"
+    )]
     pub timestamp: Timestamp,
```
rasendubi, line 13: minor: we have `serde_with` in dependencies, so we can let it do the conversions:
```suggestion
#[serde_with::serde_as]
#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
pub(super) struct Event {
    pub uuid: uuid::Uuid,
    #[serde_as(as = "serde_with::TimestampMilliSeconds")]
    pub timestamp: Timestamp,
```

rasendubi (approved):

rasendubi:
@felipecsl @petzel Is there any reason we do numeric timestamps for event ingestion instead of strings as the rest of APIs?

felipecsl:
> @felipecsl @petzel Is there any reason we do numeric timestamps for event ingestion instead of strings as the rest of APIs?

I can't recall any particular reason other than the fact that this navigates through multiple layers in the backend and having an epoch milliseconds number is easier to parse across multiple languages and frameworks I suppose.

eppo_core/src/event_ingestion/event.rs
```diff
@@ -1,11 +1,18 @@
-use crate::timestamp::Timestamp;
 use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+
+use crate::timestamp::Timestamp;
 
+#[serde_as]
```
rasendubi, line 6: We don't need `serde_as` if we don't use `serde_as` attribute below (we're using `serde(with)`)
```suggestion
```

eppo_core/src/event_ingestion/event.rs
```diff
@@ -1,11 +1,18 @@
...
+
+    #[serde(with = "chrono::serde::ts_milliseconds")]
     pub timestamp: Timestamp,
+
     #[serde(rename = "type")]
     pub event_type: String,
+
     pub payload: serde_json::Value,
-}
+}
```
rasendubi, line 18: nit: please keep newlines at the end of files. Otherwise, git is detecting the last line as changes if we add new code after it (which results in dirty git history, which may hurt some git archeology in the future)

felipecsl: was unintentional 👍🏽

PR #204, opened 2025-02-06
chore(ruby): Bump eppo_core version for next release

felipecsl:
I'm not sure why the CI version bump didn't work this time, had to manually bump it https://github.com/Eppo-exp/eppo-multiplatform/actions/runs/13188399515/job/36816113532


aarsilv (approved):

rasendubi:
The reason it's a two-step process is because `eppo_core` hash is only known after its release, so we can't update the lock file as we bump versions and it needs to be applied after `eppo_core` release

PR #208, opened 2025-02-17
Elixir SDK

schmit:
# Add Proof of Concept (PoC) for Elixir SDK Feature Flag Evaluation  

This PR introduces a Proof of Concept (PoC) for an Elixir Software Development Kit (SDK) that evaluates feature flags within the Eppo multiplatform project. The implementation includes core functionality for feature flag evaluation along with Unit Functional Coverage (UFC) test cases, all of which are passing successfully.

## Key Features  

- **Feature Flag Evaluation** – Implements core logic to assess feature flags, enabling dynamic feature management in Elixir applications.  
- **Comprehensive Testing** – Introduces UFC test cases to ensure the reliability and accuracy of the feature flag evaluation process.  

## Installation  

To install and use this Elixir SDK PoC, follow these steps:  

1. Clone the repository and check out this branch:  
   ```sh
   git clone https://github.com/Eppo-exp/eppo-multiplatform.git
   cd eppo-multiplatform
   git checkout sps/elixir-sdk
   ```

2. Ensure you have Elixir installed. If not, install it using:  
   ```sh
   brew install elixir  # macOS
   ```

3. Fetch dependencies:  
   ```sh
   mix deps.get
   ```

## Usage  

For testing locally in a script, you can instantiate the SDK using:

```
api_key = "insert api key"

string_flag_key = "new-homepage-test"

config = %EppoSdk.Client.Config{
  api_key: api_key,
  assignment_logger: EppoSdk.AssignmentLogger,
  is_graceful_mode: true,
  poll_interval_seconds: 30,
  poll_jitter_seconds: 3
}

{:ok, _pid} = EppoSdk.Server.start_link(config)
client = EppoSdk.Server.get_instance()

# wait for client to instantiate
Process.sleep(1000)

string_assignment =
  EppoSdk.Client.get_string_assignment(
    client,
    string_flag_key,
    "alice",
    %{"country" => "US", "age" => 25},
    "missing"
  )
```

## Testing  

To run the test suite, execute:  

```sh
mix test
```

This will run all test cases: the majority of tests use the cross-SDK test data to ensure consistency.

## Documentation

Documentation can be generated locally using 
```sh
mix docs
```

## Status of PR

- [x] UFC test cases are added and all pass.
- [x] Can be configured as a singleton using a GenServer
- [x] Documentation is auto-generated
- [x] Github workflow for automated testing


elixir-sdk/native/sdk_core/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+fn get_boolean_assignment<'a>(
+    env: Env<'a>,
+    flag_key: String,
+    subject_key: String,
+    subject_attributes: Term<'a>
+) -> NifResult<Term<'a>> {
+    let eppo_attributes = convert_attributes(subject_attributes)?;
+    let assignment = get_assignment(flag_key, subject_key, eppo_attributes, VariationType::Boolean);
+    match assignment {
+        Ok(Some(assignment)) => match assignment.value.as_str() {
```
rasendubi, line 186: > 2) How does assignment logging work; the logger is defined in the host language (Elixir), but some of the event information never seem to make it there (e.g. which allocation). How does this work in other SDKs? I couldn’t quite figure out how the other SDKs handle it

Yep you're right that we don't handle assignment logger inside the core and it's for individual SDKs to handle. (The main reason for that is that it's for some host languages, it's harder to call into host language from rust and we can't hold onto host language value. e.g., Ruby. So I couldn't figure out a helpful abstraction that would work for all)

The core makes it very easy to implement though. [`Assignment`](https://github.com/Eppo-exp/eppo-multiplatform/blob/21526f774f7cd03fdb2d2a29e4a6321a76967eb0/eppo_core/src/ufc/assignment.rs#L12-L20) contains a value and an optional event for logging. The value we return to the user, and the event we pass to event logger.

Tagging this place because `event` is ignored here but we want to pass it to the host language

elixir-sdk/test/sdk_core_test.exs
```diff
@@ -0,0 +1,49 @@
+defmodule SdkCoreTest do
+  use ExUnit.Case
+
+  test "init with valid config succeeds" do
+    config = %SdkCore.Config{
+      api_key: "test-key",
+      base_url: "https://api.eppo.cloud",
```
rasendubi, line 7: > 3) How do you mock a configuration so it is possible to write more tests locally?

We have a [mock-server](https://github.com/Eppo-exp/eppo-multiplatform/blob/21526f774f7cd03fdb2d2a29e4a6321a76967eb0/mock-server/) implementation that exposes all configurations from sdk-test-data on different URLs.

You can start it with [`npm run start-mock-server`](https://github.com/Eppo-exp/eppo-multiplatform/blob/21526f774f7cd03fdb2d2a29e4a6321a76967eb0/package.json#L12) and pass different base urls to the client:

- `http://127.0.0.1:8378/ufc/api`
- `http://127.0.0.1:8378/obfuscated/api`
- `http://127.0.0.1:8378/bandit/api`

rasendubi, line 7: > 1) I got get_string_assignment to work locally in a script (although the targeting rules don’t seem to be working; I got stuck trying to figure out how to debug).

For debugging, I think getting logs working is most important as they'll tell you what's going on under the hood and maybe even what's wrong.

The second tip for debugging is trying evaluation details methods — they give quite a bit of insight on why targeting didn't work (e.g., whether it's missing configuration, rule condition failed, etc. In multiplatform, it even shows individual attribute values that it considered, so if you mess up elixir-rust conversion, it'll probably show it)

For this particular failure, I think base url might be to blame — it's missing trailing `/api` here and the default value is just an empty string, so not correct either. So what I think is happening is that it can't fetch configuration and is returning default values.

elixir-sdk/native/sdk_core/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+    
+    // Convert attributes to the required format
+    Ok(Arc::new(attributes
+        .into_iter()
+        .map(|(k, v)| (Str::new(k), AttributeValue::categorical(v)))
+        .collect::<HashMap<Str, AttributeValue>>()))
+}
+
+#[rustler::nif]
+fn get_string_assignment<'a>(
```
rasendubi, line 159: > 4) How do you think about types and which functions to call (in particular, I implemented get_string_assignment as the interface, but could also implement the more generic get_assignment and then keep the get_X_assignment only on the Elixir side — but that seemed more tricky with typing

I think that depends on the host language and what's easier to do.

In Ruby, we have a single `get_assignment` function on Rust side and Ruby side implements `get_X_assignment` methods using that one function. This works well because Ruby is dynamically typed, so it's easy to return different types from Rust.

As another example, [Dart](https://github.com/Eppo-exp/eppo-multiplatform/pull/197) is statically typed, so Dart wants to know exactly what Rust functions return. So I opted for having all `get_X_assignment` functions in Rust + wrappers in Dart.

So I think the answer is: do what's easier in Elixir

elixir-sdk/README.md
rasendubi, line 1: > My main question is, given this steel thread, how much more work is there to do? I think there’s quite a lot (I’m guessing somewhere between 5x and 20x) and wanted to understand whether it is feasible for me to keep working on it at random hours and make meaningful progress.

I actually think that you already implemented the hardest part — figuring out how to interact between the languages. After you get the targeting bug resolved (which I hope is easy), what's left is:
- get assignment logger work (shall be quite trivial as well once you pass event to elixir side)
- define the rest of API and write all the boilerplate code (not familiar with elixir but it'll probably take between a day and three if you want to get fancy). Bandits with their "context attributes" may get tricky but maybe not given that Elixir seems loosely typed. (e.g., for Dart, it felt awkward to have generic attributes and context attributes as different types, so I found a way to unify them into a single attribute type)
- packaging and releasing (very much depends on the tooling if that's easy or not)

elixir-sdk/native/sdk_core/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+    let mut instance = CLIENT_INSTANCE
+        .write()
+        .map_err(|e| format!("Failed to acquire write lock: {}", e))?;
+    
+    if let Some(existing) = instance.take() {
+        // Shutdown existing client
+        drop(existing);
+    }
+    
+    *instance = Some(client.clone());
```
rasendubi, line 99: minor: if you can, I would like to push singleton up as much as possible. Ideally, having a non-singleton EppoClient "class" in Elixir with a thin wrapper to initialize/get instance.

This shall make it easier to support multiple clients in the future. Not sure about Elixir's threading model but that may also help us to get rid of locks (most of core is designed to allow concurrent access from multiple threads). Non-singletons are also kind of easier to reason about

schmit: Hmm Elixir is functional and hence the singleton pattern does not make as much sense; hence I figured it would be easiest to completely hide the client object in Elixir. There's something called a GenServer we could try using, but I feel like it'd needlessly complicate things (and you can only have 1 with a given name, so wouldn't support multiple clients easily either). 

There's probably a good option available, I just don't know it

elixir-sdk/native/sdk_core/src/lib.rs
schmit: Great, Elixir is like Ruby; I will try to understand the Ruby client better and try that approach -- I had trouble figuring out how to return a generic value from Rust with the rustler library.

eppo_core/src/ufc/models.rs
```diff
@@ -96,7 +96,7 @@ pub(crate) struct FlagWire {
 }
 
 /// Type of the variation.
-#[derive(Debug, Serialize, Deserialize, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Serialize, Deserialize, Clone, Copy, PartialEq, Eq, rustler::NifUnitEnum)]
```
schmit, line 99: @rasendubi it feels like a bit of a smell to add this but it seemed like the easiest way to make sure we can convert Elixir atoms (such as `:string` to a VariationType). Let me know if you want me to avoid adding rustler and this derive in eppo_core

elixir-sdk/native/sdk_core/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+        let json_value = serde_json::to_value(&event)
+            .map_err(|e| rustler::Error::Term(Box::new(format!("Failed to serialize event: {:?}", e))))?;
+        if let serde_json::Value::Object(map) = json_value {
+            let converted: HashMap<String, String> = map.into_iter()
+                .map(|(k, v)| (k, v.to_string()))
+                .collect();
+            Ok(converted.encode(env))
+        } else {
+            Err(rustler::Error::Term(Box::new("Event did not serialize to an object".to_string())))
+        }
```
schmit, line 181: This is pretty gross -- I wonder what's best here, perhaps just send the json as a string and then convert to an object in Elixir (losing the performance of Rust)? Or maybe just keep it as a string (because we'll be logging a string anyway), that seems gross too

eppo_core/src/ufc/models.rs
rasendubi: I think that's ok'ish. We have some derives for python/ruby in core as well. Unfortunately, because of rust impl rules, we can't add these implementation outside of core so I think that's fine for now.

The only thing is that this should depend on a feature flag, so we don't pull any of elixir libraries into other languages SDKs.

```suggestion
#[derive(Debug, Serialize, Deserialize, Clone, Copy, PartialEq, Eq)]
#[cfg_attr(feature = "rustler", derive(rustler::NifUnitEnum))]
```

elixir-sdk/native/sdk_core/src/lib.rs
rasendubi: Have you looked into serde_rustler? (I didn't really but it looks like something that should help converting between rust/elixir given we already have Serialize/Deserialize implementations)

schmit: Good call, the repo is archived and hasn't been updated; also doesn't seem to be working with more recent Elixir versions. Sounds like there were some plans to merge this into Rustler but the issue is still open: https://github.com/rusterlium/rustler/issues/200

elixir-sdk/native/sdk_core/src/lib.rs
schmit: Now updated to use the generic function (a la Ruby)

elixir-sdk/native/sdk_core/src/lib.rs
schmit: This was super easy to implement, love your thoughtfulness in implementing the core :)

elixir-sdk/native/sdk_core/src/lib.rs
rasendubi: oh, that sucks. it looks like there's some experimental serde support in rustler already: https://docs.rs/rustler/latest/rustler/serde/index.html

We may write our own encoder from events to elixir terms. Shouldn't be hard — just boring

But I think to json -> from json approach is also viable for the first version.

I think another option is keeping event in a resource arc, so the user can decide whether they want to serialize it to string or query data from it

eppo_core/src/ufc/models.rs
schmit: I couldn't get this to work yet, might need some extra eyes later.

elixir-sdk/native/sdk_core/src/lib.rs
schmit: For now, opted to send raw json to Elixir, and convert using Elixir's JSON library. Helps with nesting too. We can optimize later.

schmit:
Current status: got get_X_assignment and get_X_assignment_details to work locally. Up next:

- [ ] Write tests using sdk-test-data
- [ ] Create some examples (script, Phoenix app)

elixir-sdk/native/sdk_core/src/lib.rs
schmit: Fixed this using a GenServer, actually is pretty nice, hope you like it!

elixir-sdk/lib/assignment_logger.ex
```diff
@@ -0,0 +1,12 @@
+defmodule Eppo.AssignmentLogger do
```
schmit, line 1: Unsure whether to use `Eppo.X` or `EppoSDK.X`
```suggestion
defmodule EppoSDK.AssignmentLogger do
```

Curious what other think

elixir-sdk/lib/client.ex
```diff
@@ -0,0 +1,461 @@
+defmodule Eppo.Client do
```
schmit, line 1: This module is the main interface for the client

elixir-sdk/lib/server.ex
```diff
@@ -0,0 +1,72 @@
...
+  """
+  def get_instance() do
+    case GenServer.whereis(__MODULE__) do
+      nil -> raise "Eppo client not initialized. Ensure Eppo.Client.Server is started."
+      pid -> GenServer.call(pid, :get_client)
+    end
+  end
+
+  @impl true
+  def init(%Eppo.Client.Config{} = config) do
```
schmit, line 61: Alternatively, we could require passing a client rather than a config?

elixir-sdk/lib/assignment_logger.ex
schmit: Defaulting to call it `EppoSdk` for now, seems most consistent but open to other suggestions

elixir-sdk/native/sdk_core/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+        sdk_metadata: SDK_METADATA,
+    });
+
+    let client = ResourceArc::new(EppoClient {
+        evaluator,
+        background_thread,
+    });
+
+    Ok(client)
+}
```
felipecsl, line 89: it seems like we have this same `init` routine repeated 3 times for each SDK now, might be a good candidate for moving to eppo-core

felipecsl (commented):
Awesome work. I'm not really familiar with Elixir, but on a high level looks good to me.

felipecsl:
I didn't approve yet since it seems like you're still iterating on it and it's probably a good idea getting another set of eyes, but no blockers from me

leoromanovsky:
🚀 

* Please demonstrate a working local program with the elixir sdk integrated
* If you could avoid force pushing after starting a review, that would be appreciated to preserve the timeline of conversations, especially with the great discussion happening here. I am liberal with force pushes until reviews start and afterwards run `git merge main` to catch your branch up.

schmit:
@leoromanovsky great feedback!

I currently have a script locally that demonstrates everything works, I was planning to add that script along with a simple Phoenix example app in a follow-up PR so that I can also test/show installing the SDK from the Hex package manager. But if you want I can add the script now.

And I hear you re: force pushes, will do 🙇

elixir-sdk/test/sdk_core_test.exs
schmit: Just a quick update here that this now works, the conversion of subject_attributes needed more work.

leoromanovsky:
> @leoromanovsky great feedback!
> 
> I currently have a script locally that demonstrates everything works, I was planning to add that script along with a simple Phoenix example app in a follow-up PR so that I can also test/show installing the SDK from the Hex package manager. But if you want I can add the script now.
> 
> And I hear you re: force pushes, will do 🙇

From my point of view, this is a tremendous among of code to review visually. If we approve the PR how do we know its working and ready for customers? 

Please include at least:

* a testing section showing how you compile locally + showing it running inside a small app
* add sdk test data testing at the library level

schmit:
> From my point of view, this is a tremendous among of code to review visually. If we approve the PR how do we know its working and ready for customers?

At this point, all the cross-platform SDK test cases are passing, which indicates to me that the logic works correctly (and if not, we should create more test cases). 

The thing I haven't been able to test is to install the SDK from Hex (the package manager) directly; once this is merged and published I will create a follow-up PR that adds examples and ensures that works.
 
> Please include at least:
> 
> * a testing section showing how you compile locally + showing it running inside a small app

This is buried in all the files but is explained in `README.md` -- this is also added to the automatically generated documentation. Run `mix docs` to generate them; these should be available on [hex.docs](https://hexdocs.pm/) once the package is published.

I have also updated the PR description with a simple script to show how the SDK works.

> * add sdk test data testing at the library level

What additional tests beyond the cross-platform SDK test cases would you like to see?

elixir-sdk/lib/core.ex
```diff
@@ -0,0 +1,37 @@
+defmodule EppoSdk.Core do
```
schmit, line 1: This module just defines the Rust interface -- it was confusing to me at first, but all these functions are "replaced" by Rust functions with the same names.

eppo_core/src/ufc/models.rs
leoromanovsky: ✨ https://github.com/Eppo-exp/eppo-multiplatform/pull/215

leoromanovsky: Since you are modifying `eppo_core`, it will need to be published and that needs a changeset: https://github.com/Eppo-exp/eppo-multiplatform/pull/216

You can also try the tool locally with `yarn changeset`, it has a guided demo

leoromanovsky (approved):

elixir-sdk/native/sdk_core/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+authors = []
+edition = "2021"
+
+[lib]
+name = "sdk_core"
+crate-type = ["cdylib"]
+
+[dependencies]
+rustler = "0.36.1"
+eppo_core = { path = "../../../eppo_core", features = ["rustler"] }
```
leoromanovsky, line 13: this needs an actual dependency, you can override like this for local testing: https://github.com/Eppo-exp/eppo-multiplatform/pull/197#issuecomment-2680522104

just bumped `eppo_core` to `9.0.0`

.changeset/light-socks-wait.md
```diff
@@ -0,0 +1,5 @@
+---
+"eppo_core": patch
```
rasendubi, line 2:
```suggestion
"eppo_core": minor
```

elixir-sdk/native/sdk_core/README.md
rasendubi, line 1: minor: this file seems to be an outdated boilerplate. Do we still need it?

elixir-sdk/native/sdk_core/src/conversion.rs
```diff
@@ -0,0 +1,56 @@
...
+            // Integers are converted to f64 and stored as numeric.
+            AttributeValue::numeric(i as f64)
+        } else if let Ok(f) = value_term.decode::<f64>() {
+            AttributeValue::numeric(f)
+        } else if let Ok(s) = value_term.decode::<String>() {
+            // Strings are stored as categorical attributes.
+            AttributeValue::categorical(s)
+        } else {
+            // If none of the supported types matched, return a null attribute.
+            AttributeValue::null()
```
rasendubi, line 28: minor: shall we warn?

elixir-sdk/native/sdk_core/src/conversion.rs
```diff
@@ -0,0 +1,56 @@
+use rustler::{Encoder, Env, NifResult, Term};
+use std::collections::HashMap;
+use std::sync::Arc;
+use eppo_core::{AttributeValue, Str, events::AssignmentEvent, ufc::AssignmentValue};
+use rustler::types::atom;
+
+
+pub fn convert_attributes(subject_attributes: Term) -> NifResult<Arc<HashMap<Str, AttributeValue>>> {
+    // Obtain an iterator over the map's key-value pairs.
+    let map: HashMap<String, Term> = subject_attributes.decode()?;
```
rasendubi, line 10: minor: would it make sense to use [`MapIterator`](https://docs.rs/rustler/latest/rustler/types/struct.MapIterator.html) here? Very minor, but `HashMap` allocates the structure and string keys ahead of time which then get discarded. It looks like we can use map iterator, so we don't need additional interim allocation

Another option is implementing `Decoder` for `Str` and `AttributeValue`, so we automatically get efficient implementation of `Decoder` for `HashMap<Str, AttributeValue>`

elixir-sdk/native/sdk_core/src/conversion.rs
```diff
@@ -0,0 +1,56 @@
...
+            // Booleans are stored as categorical attributes.
+            AttributeValue::categorical(b)
+        } else if let Ok(i) = value_term.decode::<i64>() {
+            // Integers are converted to f64 and stored as numeric.
+            AttributeValue::numeric(i as f64)
+        } else if let Ok(f) = value_term.decode::<f64>() {
+            AttributeValue::numeric(f)
+        } else if let Ok(s) = value_term.decode::<String>() {
+            // Strings are stored as categorical attributes.
+            AttributeValue::categorical(s)
```
rasendubi, line 25: minor: it looks like `&str` implements `Decode`, and its very likely to be more efficient than allocating a new `String`. `Str` (and thus `AttributeValue`) also has a couple of optimizations for small strings to avoid copying, so `&X -> &str -> Str` is more preferable than `&X -> String -> Str`

```suggestion
        } else if let Ok(s) = value_term.decode::<&str>() {
            // Strings are stored as categorical attributes.
            AttributeValue::categorical(s)
```

Same also goes for map keys

elixir-sdk/native/sdk_core/src/conversion.rs
```diff
@@ -0,0 +1,56 @@
...
+}
+
+pub fn convert_event_term<'a>(env: Env<'a>, event: Option<AssignmentEvent>) -> NifResult<Term<'a>> {
+    if let Some(event) = event {
+        let json_value = serde_json::to_value(&event)
+            .map_err(|e| rustler::Error::Term(Box::new(format!("Failed to serialize event: {:?}", e))))?;
+        Ok(json_value.to_string().encode(env))
+    } else {
+        Ok(atom::nil().encode(env))
+    }
```
rasendubi, line 55: minor: is [`SerdeTerm`](https://docs.rs/rustler/latest/rustler/serde/struct.SerdeTerm.html) something that would be useful here?

I think it works like this:
```suggestion
    Ok(SerdeTerm(&event).encode(env))
```

elixir-sdk/native/sdk_core/src/lib.rs
rasendubi: @felipecsl this somewhat makes sense but not all SDKs share this init. Dart and Rust SDKs don't spawn `BackgroundThread` in init, so that's a 3/5 kind of reuse—it seems the same but isn't, it just happens to be the same in a couple of SDKs

elixir-sdk/native/sdk_core/src/conversion.rs
rasendubi, line 1: minor: given we already have rustler-specific code in `eppo_core`, consider implementing `Encoder`/`Decoder` traits. This should make the code simpler but also more efficient: if we implement `Decoder` for `Str` and `AttributeValue`, we'll be able to decode directly to `HashMap<Str, AttributeValue>` without intermediate `HashMap<String, Term>` allocation

rasendubi (approved):
great job!

elixir-sdk/native/sdk_core/src/conversion.rs
schmit: Fantastic, this works!

elixir-sdk/native/sdk_core/src/conversion.rs
schmit: Sweet!

elixir-sdk/native/sdk_core/src/conversion.rs
schmit: fantastic :bow:

PR #212, opened 2025-02-21
fix(python): Options that are None are also None in python

bennettandrews:
## Description

Options that are `None` in rust are also `None` in python instead of empty tuples. 

Currently when it is converted back to python it ends up as tuple `()`  which will create an empty tuple object instead. 

## Testing

Updated test to include checking conversion. 

## References 

fixes: https://github.com/Eppo-exp/eppo-multiplatform/issues/211

### DIFF
--- a/.changeset/cyan-socks-fold.md
+++ b/.changeset/cyan-socks-fold.md
@@ -1,8 +1,8 @@
 ---
 "eppo_core": patch
 "python-sdk": patch
+"ruby-sdk": patch
+"rust-sdk": patch
 ---
 
-fix(python): Options that are None are also None in python
-
-None Attributes are now correctly converted to Python's `None` instead of `()` empty tuple.
+Fix `AttributeValue` serialization, so `Null` attributes are properly serialized as None instead of unit value.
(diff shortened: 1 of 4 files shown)


rasendubi:
Thanks for debugging and fixing this! Extra kudos for adding tests ❤️

rasendubi:
Hm... it looks like the test is failing:
```
>       assert event["subjectAttributes"]["empty"] is None
E       assert () is None
```

Did you get it passing locally?

rasendubi:
Ok, I figured it out. The bug is coming from `AttributeValue` having bad implementation for `serde::Serialize`. I've fixed it and added a commit to your branch

bennettandrews:
@rasendubi Thanks. I was still working on figuring out some pieces of getting local builds to work so I wasn't able to test this yet.

leoromanovsky (approved):
Great find thank you! The other CI check is failing and will be fixed in the MSRV pr here (https://github.com/Eppo-exp/eppo-multiplatform/pull/213).

We'll get this packaged up and shipped for you early next week - cheers!

PR #222, opened 2025-02-28
chore(ci): dart-sdk rust cross compile for android target

felipecsl:


felipecsl:
`flutter_rust_bridge_codegen` is failing here, gotta figure out why

felipecsl:
YES this is FINALLY ready to go, PTAL @rasendubi @leoromanovsky

.github/workflows/android.yml
```diff
@@ -0,0 +1,132 @@
...
+      matrix:
+        target:
+          [aarch64-linux-android, x86_64-linux-android, armv7-linux-androideabi]
+        include:
+          - target: aarch64-linux-android
+            android_abi: arm64-v8a
+          - target: x86_64-linux-android
+            android_abi: x86_64
+          - target: armv7-linux-androideabi
+            android_abi: armeabi-v7a
```
leoromanovsky, line 41: https://developer.android.com/ndk/guides/abis

Could you try to add `x86` 32-bit and see if it compiles? That will cover all our bases.

leoromanovsky: This covers all major platforms - really awesome to be able to deliver this that feels every bit as comprehensive as native code

.github/workflows/android.yml
```diff
@@ -0,0 +1,134 @@
...
+        with:
+          workspaces: dart-sdk/rust -> target
+          key: ${{ matrix.target }}
+
+      - name: Override eppo_core for testing
+        shell: bash
+        run: |
+          mkdir -p "${CARGO_HOME:-$HOME/.cargo}"
+          echo "[patch.crates-io.eppo_core]" >> "${CARGO_HOME:-$HOME/.cargo}/config.toml"
+          echo "path = '${{ github.workspace }}/eppo_core'" >> "${CARGO_HOME:-$HOME/.cargo}/config.toml"
```
leoromanovsky, line 71: does this need to be removed or later after publishing?

dart-sdk/rust/Cross.toml
```diff
@@ -0,0 +1,35 @@
...
+volumes = [ "SDK_DIR=../", ]
+
+[target.aarch64-linux-android]
+image = "ghcr.io/cross-rs/aarch64-linux-android:main"
+
+[target.x86_64-linux-android]
+image = "ghcr.io/cross-rs/x86_64-linux-android:main"
+
+[target.armv7-linux-androideabi]
+image = "ghcr.io/cross-rs/armv7-linux-androideabi:main"
```
leoromanovsky, line 14: x86 32-bit target needed here?

leoromanovsky (approved):

.github/workflows/android.yml
felipecsl: yeah I think for CI here we want to build against the current unreleased version of eppo_core, but for publishing (assuming that will be in a separate github workflow) we'll want to use the released version instead. So yeah I think this is fine as is, cc @rasendubi to keep me honest

dart-sdk/rust/Cross.toml
felipecsl: these are actually the default images so this config is probably unnecessary. I'm debading whether to keep this or remove, might remove for simplicity

.github/workflows/android.yml
rasendubi: oh, yeah. good catch that we need to remove publish against released version of eppo_core

I've updated CI here: https://github.com/Eppo-exp/eppo-multiplatform/pull/229

PR #228, opened 2025-03-04
feat(dart-sdk): Add optional log level argument

felipecsl:


dart-sdk/rust/src/api/client.rs
```diff
@@ -32,20 +32,31 @@ impl CoreClient {
...
         #[frb(default = "3000")] poll_jitter_ms: u64,
+        #[frb(default = "info")] log_level: String,
     ) -> CoreClient {
-        log::set_max_level(log::LevelFilter::Info);
+        log::set_max_level(log::LevelFilter::Debug);
 
-        let base_url = base_url.unwrap_or_else(|| DEFAULT_BASE_URL.to_owned());
+        {
+            let mut builder = env_logger::Builder::from_env(
+                env_logger::Env::new().write_style("EPPO_LOG_STYLE"),
```
leoromanovsky, line 41: Is this a dart/flutter canonical environment style for this kind of parameter? Just asking if a customer sees this will they "get it" or better yet is there a name that will "just work" without them reading a doc?

leoromanovsky (approved):

dart-sdk/rust/src/api/client.rs
felipecsl: I'm not sure. I took this partially from the ruby-sdk client https://github.com/Eppo-exp/eppo-multiplatform/blob/main/ruby-sdk/ext/eppo_client/src/client.rs#L77-L81
however, I haven't figured out a way to pass environment variables from flutter all the way down into the rust native code, not sure if it's even possible, so we may want to remove this `EPPO_LOG_STYLE` - as usual, I'l let @rasendubi chime in

dart-sdk/rust/src/api/client.rs
```diff
@@ -32,20 +32,31 @@ impl CoreClient {
...
         #[frb(default = "30000")] poll_interval_ms: u64,
         #[frb(default = "3000")] poll_jitter_ms: u64,
+        #[frb(default = "info")] log_level: String,
     ) -> CoreClient {
-        log::set_max_level(log::LevelFilter::Info);
+        log::set_max_level(log::LevelFilter::Debug);
 
-        let base_url = base_url.unwrap_or_else(|| DEFAULT_BASE_URL.to_owned());
+        {
+            let mut builder = env_logger::Builder::from_env(
```
rasendubi, line 40: We have another logger setup from via default utils: https://github.com/Eppo-exp/eppo-multiplatform/blob/b6ede436e1a9ea9657d7f2fc158328d9c3af59fc/dart-sdk/rust/src/api/mod.rs#L61

From what I get, it integrates better with whatever platform is using for logging. And env logger will likely conflict with it.

I think we just need to pass the desired level to `set_max_level` and that should be enough?

rasendubi:
> 🔔 Note: Right now, integrating the dart-sdk into a flutter app requires manually running flutter_rust_bridge_codegen generate --config-file flutter_rust_bridge.yaml, which is obviously not great. I think we should publish the generated sources with the dart-sdk, but that's just an aside here.

We're running this step before the release and including generated files in the release tarball, so you don't need when installing SDK from pub.dev

felipecsl:
> > 🔔 Note: Right now, integrating the dart-sdk into a flutter app requires manually running flutter_rust_bridge_codegen generate --config-file flutter_rust_bridge.yaml, which is obviously not great. I think we should publish the generated sources with the dart-sdk, but that's just an aside here.
> 
> We're running this step before the release and including generated files in the release tarball, so you don't need when installing SDK from pub.dev

Oh nice I didn't realize that, awesome!

PR #249, opened 2025-03-19
[elixir-sdk] Add wait-for-initialization

schmit:
The Elixir equivalent of https://github.com/Eppo-exp/eppo-multiplatform/pull/247

### DIFF
--- a/elixir-sdk/native/sdk_core/src/lib.rs
+++ b/elixir-sdk/native/sdk_core/src/lib.rs
@@ -30,7 +30,7 @@ const SDK_METADATA: SdkMetadata = SdkMetadata {
 pub struct EppoClient {
     pub evaluator: Evaluator,
     pub background_thread: BackgroundThread,
-    pub configuration_poller: Option<ConfigurationPoller>,
+    pub configuration_poller: ConfigurationPoller,
 }
 
 #[rustler::resource_impl]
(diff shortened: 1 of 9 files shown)


elixir-sdk/native/sdk_core/src/lib.rs
```diff
@@ -29,6 +30,7 @@ const SDK_METADATA: SdkMetadata = SdkMetadata {
 pub struct EppoClient {
     pub evaluator: Evaluator,
     pub background_thread: BackgroundThread,
+    pub configuration_poller: Option<ConfigurationPoller>,
```
rasendubi, line 33: Given we always have a poller, does this field needs to be an `Option`?

related: do these fields need to be public? Just want to confirm that they are not part of our public API and we can change them later

rasendubi (approved):

elixir-sdk/native/sdk_core/src/lib.rs
```diff
@@ -151,4 +154,29 @@ fn get_assignment_details<'a>(
...
+        .block_on(async {
+                tokio::time::timeout(
+                    Duration::from_secs_f64(timeout_secs),
+                    client.configuration_poller.wait_for_configuration(),
+                )
+                .await
+            })
+            .inspect_err(|err| {
+                log::warn!(target: "eppo", "failed to wait for initialization: {:?}", err);
+            });
```
rasendubi, line 177: nit: the indentation is off here

PR #254, opened 2025-03-21
[elixir] Fix eppo_core package import

schmit:
Fix how to import eppo_core locally vs as a package

### DIFF
--- a/elixir-sdk/.cargo/config.toml
+++ b/elixir-sdk/.cargo/config.toml
@@ -1,0 +1,5 @@
+[target.'cfg(target_os = "macos")']
+rustflags = [
+    "-C", "link-arg=-undefined",
+    "-C", "link-arg=dynamic_lookup",
+]
--- a/elixir-sdk/.cargo/config.toml
+++ b/elixir-sdk/.cargo/config.toml
@@ -1,5 +1,0 @@
-[target.'cfg(target_os = "macos")']
-rustflags = [
-    "-C", "link-arg=-undefined",
-    "-C", "link-arg=dynamic_lookup",
-]
--- a/elixir-sdk/.cargo/config.toml.template
+++ b/elixir-sdk/.cargo/config.toml.template
@@ -1,0 +1,11 @@
+# Copy this file to .cargo/config.toml (without the .template extension)
+# to use the local eppo_core package during development.
+
+[target.'cfg(target_os = "macos")']
+rustflags = [
+    "-C", "link-arg=-undefined",
+    "-C", "link-arg=dynamic_lookup",
+]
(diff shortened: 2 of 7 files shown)


elixir-sdk/.cargo/config.toml
```diff
@@ -1,5 +0,0 @@
-[target.'cfg(target_os = "macos")']
```
schmit, line 1: This file shouldn't be checked in, hence removing it

elixir-sdk/native/sdk_core/Cargo.toml
```diff
@@ -10,7 +10,7 @@ crate-type = ["cdylib"]
 
 [dependencies]
 rustler = { version = "0.36.1", features = ["serde"] }
-eppo_core = { version = "=9.1.1", path = "../../../eppo_core", features = ["rustler"] }
+eppo_core = { version = "9.1.1", features = ["rustler"] }
```
schmit, line 13: main fix

elixir-sdk/.cargo/config.toml.template
```diff
@@ -0,0 +1,11 @@
...
+# to use the local eppo_core package during development.
+
+[target.'cfg(target_os = "macos")']
+rustflags = [
+    "-C", "link-arg=-undefined",
+    "-C", "link-arg=dynamic_lookup",
+]
+
+[patch.crates-io]
+eppo_core = { path = "../../eppo_core" }
```
schmit, line 11: this overrides the import of eppo_core to use the local package

elixir-sdk/native/sdk_core/Cargo.toml
rasendubi, line 13: I believe we still want to pin a specific version of core. Otherwise, it technically may ship unexpected changes for users who build from source
```suggestion
eppo_core = { version = "=9.1.1", features = ["rustler"] }
```

rasendubi (approved):

elixir-sdk/.cargo/config.toml.template
rasendubi: isn't this too much `..`? I believe it's counting starting from `elixir-sdk`
```suggestion
eppo_core = { path = "../eppo_core" } 
```

schmit: Yes good eyes! I am certain I fixed this before but clearly not 🤔  -- updated.

elixir-sdk/native/sdk_core/Cargo.toml
schmit: Fixed!

PR #275, opened 2025-04-22
chore: make dependabot update nested lock files as well

rasendubi:
Found that dependabot wasn't bumping Ruby SDK lockfile. Adding ruby-sdk directory explicitly should fix that.

Note that this does remove root directory, so the root lockfile will not get updated. That should be fine as it is unused for releases (and we should probably remove it altogether libraries shouldn't normally have lockfiles committed).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,10 +11,14 @@ updates:
       interval: "daily"
 
   - package-ecosystem: cargo
-    directory: "/"
+    directories:
+      - "/eppo_core"
+      - "/rust-sdk"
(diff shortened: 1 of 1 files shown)


typotter (approved):
Any idea why it wasn't seeing the ruby-sdk lockfile?

rasendubi:
@typotter Dependabot was only pointed to `/` root workspace. It was traversing to and updating member packages correctly. However, `/ruby-sdk` is its own workspace (and not listed as a member of root workspace), and thus it wasn't traversed and its lock file wasn't updated

PR #424, opened 2026-07-22
ci: fix version workflow

dd-oleksii:
version workflow is failing with the following error:

    > ruby-sdk@3.8.1 postversion
    > ./postversion.sh

    + jq -r .version ./package.json
    + VERSION=3.8.1
    + cargo set-version -p eppo_client 3.8.1
    Error: `cargo metadata` exited with an error:     Updating crates.io index
    error: failed to select a version for the requirement `eppo_core = "=10.0.1"`
    candidate versions found which didn't match: 10.0.0, 9.3.0, 9.2.0, ...

This is likely caused by cargo not finding the eppo_core inside ruby workspace.

Set the eppo_core override in ~/.cargo/config.toml.

Additional fixes:
- pin cargo-edit to a specific version
- replace sed's -i option with manual handling of tmp file to make it portable across linux and macos

### DIFF
--- a/.github/workflows/version.yml
+++ b/.github/workflows/version.yml
@@ -38,7 +38,14 @@ jobs:
         run: rustup update stable && rustup default stable
 
       - name: Install cargo-edit
-        run: cargo install cargo-edit
+        run: cargo install cargo-edit@0.13.13
+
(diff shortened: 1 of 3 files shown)


aarsilv (approved):
👍

